PR #2, opened 2018-10-15
Add JSON schema for genes collection

jayrbolton:
Quick example of what a PR might look like for submitting a new document schema for the database.


src/schemas/genes.json5
```diff
@@ -0,0 +1,24 @@
+{
```
scanon: How is the key ID defined?  We may want to document that somehow.  

Specifically for the genome and gene, I would like to have some discussion about the key.  Are these going to be explicitly linked to the WS object?  If so, then it would be good if the key mapped to the object.  e.g.

id: WS:1/2/3 for a genome or other node that mapped directly to an object.

for feature or other sub-objects we should have some convention.  e.g. 

id: WS:1/2/3:feature/123

This roughly mirrors what is done in search.

The other question is do we have other genome-like and gene-like nodes in the graph that don't map directly to a WS object.

src/schemas/genes.json5
```diff
@@ -0,0 +1,24 @@
...
+  description: 'Schema for the `genes` collection.',
+  version: 1,
+  type: 'object',
+  required: [],
+  properties: {
+    note: {type: 'string'},
+    codon_start: {type: 'string'},
+    inference: {type: 'string'},
+    location_end: {type: 'integer', minimum: 0},
+    location_start: {type: 'integer', minimum: 0},
```
JamesJeffryes: So gene locations can be actually pretty complex because genes can be spliced together from multiple locations on the genome. The "Start" and "End" supplied by BioPython can be pretty deceptive/useless for Euks and is also significantly different from how KBase stores locations.

It's not exactly how I would have designed it but locations in KBase are lists of triples of (contig, start_location, length)

jayrbolton: The location start and end fields are directly in the source data from NCBI, so I don't think it's supplied by biopython in any case I've seen; biopython just parses it out. All the data in this prototype schema is just a direct pull from the data points in NCBI. I'm not really doing any transformation. I'll look more at the KBase workspace type definitions

src/schemas/genes.json5
jayrbolton: - good point, we should include `_key` in all schemas. FYI, JSON Schema can include metadata like "description", "default", and "examples" for every field.
- it might also be worth having a separate "workspace_ref" field for documents in the db that map to a workspace object (eg. a genome). I'm not familiar enough with the KBase workspace to know all the stuff there to know how things might map

PR #4, opened 2018-11-02
Prov specs

JamesJeffryes:
@jayrbolton Here's some initial sketches for specs for a few types. hope to get your thoughts on them

### DIFF
--- a/schemas/object_hashes.json
+++ b/schemas/object_hashes.json
@@ -1,0 +1,27 @@
+{
+  "definitions": {},
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "version": 1,
+  "additionalProperties": false,
+  "required": [
+    "_key",
+    "type"
+  ],
+  "properties": {
+    "_key": {
+      "type": "string",
+      "description": "The key is the hash",
+      "examples": [
+        "2406642b28312b3ccbfb2e17e231e2c7"
+      ]
+    },
+    "type": {
+      "type": "string",
+      "description": "The hashing algorithm used",
+      "examples": [
+        "MD5"
+      ]
+    }
+  }
+}
--- a/schemas/users.json
+++ b/schemas/users.json
@@ -1,5 +1,4 @@
 {
-  "definitions": {},
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "version": 1,
@@ -7,7 +6,6 @@
   "required": ["_key"],
   "properties": {
     "_key": {
-      "$id": "#/properties/_key",
       "type": "string",
       "description": "The username for this user",
       "examples": [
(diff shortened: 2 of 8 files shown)


schemas/object_hashes.json
```diff
@@ -0,0 +1,27 @@
+{
```
jayrbolton: Should this be a separate document? Could it be a sub-object inside other documents?
I'm just generally wary of separating out documents too much, because it's going to make the queries complicated

schemas/object_hashes.json
```diff
@@ -0,0 +1,27 @@
...
+  "definitions": {},
```
jayrbolton: We can leave out the "definitions" on these

schemas/users.json
```diff
@@ -0,0 +1,19 @@
+{
+  "definitions": {},
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "version": 1,
+  "additionalProperties": false,
+  "required": ["_key"],
+  "properties": {
+    "_key": {
+      "$id": "#/properties/_key",
```
jayrbolton: We can scrub these "$id" properties

schemas/workspaces.json
```diff
@@ -0,0 +1,35 @@
...
+      "examples": [
+        "35414"
+      ],
+      "pattern": "^\\d+$"
+    },
+    "name": {
+      "type": "string",
+      "description": "The workspace name for this workspace",
+      "examples": [
+        "jjeffryes:narrative_1534187093329"
```
jayrbolton: Can these be regexed?

schemas/ws_object_versions.json
```diff
@@ -0,0 +1,38 @@
...
+      "type": "string",
+      "description": "The UPA for this data",
+      "examples": [
+        "35414:73:1"
+      ],
+      "pattern": "^\\d+:\\d+:\\d+$"
+    },
+    "size": {
+      "type": "integer",
+      "description": "Size in bits",
```
jayrbolton: Are you sure this is bits, not bytes?

jayrbolton:
One general comment, not a request on this PR:
Down the line we could separate out some of these IDs/names if we need to. 
- For example, a type: - `{_key: "Module.Type-1.0", module_name: "Module", type: "Type", version: [1, 0]}`
- Or an object: - `{_key: "1:2:3", ws_id: 1, obj_id: 2, version: 3}`
Would save a lot of parsing, especially in aql queries. We don't need it now though.

schemas/workspaces.json
JamesJeffryes: So a WS generated by a narrative would certainly follow this pattern but there could be other sources of WS as well.

schemas/object_hashes.json
JamesJeffryes: I think we should talk to Shane about that one. I think the query that this structure neetly provides is given an object, list the identical objects. I don't know Arango well enough to know if that is better supported by an indexed field.

jayrbolton:
Going to merge this to keep things moving. If we decide to split vertices/edges into subfolder, we can make another PR

PR #8, opened 2018-11-06
Clean up various schemas

jayrbolton:
@JamesJeffryes : I went through the schemas you added and cleaned some things up. I renamed a couple edges to make things more explicit. With edge names like "contains", we want to make sure the language around the description is pretty generic


schemas/vertices/workspaces.json
```diff
@@ -31,13 +30,10 @@
       "examples": [
         "2014-04-09T19:40:25+0000"
       ],
-      "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}$"
+      "format": "date-time"
```
jayrbolton: See here: http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1

PR #10, opened 2018-11-07
Basic views

jayrbolton:
This adds a couple very basic, generic views. You can run these on any collection. I'm using these in API tests.

Other generic view ideas:
- Traverse from vert to vert via one level of edges, two levels of edges, etc.
- List all documents in a collection with an equality filter on an attribute

Idea to make views more flexible: allow flags that add or remove sections of the query (eg a filter line)

### DIFF
--- a/views/list_all_documents_in_collection.aql
+++ b/views/list_all_documents_in_collection.aql
@@ -1,0 +1,6 @@
+// Return *all* full documents in a collection
+// Args:
+//   collection - name of collection to count docs
+
+for v in @@collection
+  return v
--- a/views/count_documents_in_collection.aql
+++ b/views/count_documents_in_collection.aql
@@ -5,3 +5,4 @@
 for v in @@collection
   collect with count into length
   return length
+


views/list_all_documents_in_collection.aql
```diff
@@ -0,0 +1,6 @@
+// Return *all* full documents in a collection
+// Args:
+//   collection - name of collection to count docs
+
+for v in @@collection
+  return v
```
JamesJeffryes: We might want to put a limit on this so someone does not tie up Arango by running this on ws_object_versions. We also need to give another think on access control.

jayrbolton: The API will limit things. Query results are limited to 100 and more result pages can be fetched through a cursor. RAM usage for a single query is also capped. I think this should be fine in prod actually.

jayrbolton: But regarding access control -- yes. We'll have to update this for that.

JamesJeffryes: so `Return *all* full documents in a collection` is a LIE! Good to know about the caps on queries though.

jayrbolton: Not a lie if you follow the magic cursor!
Say you had 200 docs in a collection:
- POST /query {name: list_all_documents_in_collection} -> 100 results, cursor 123
- GET /cursor {id: 123} -> 100 next results, no cursor

jayrbolton: I might change the above API to make things a little less confusing:

* GET /query_results {view: lsit_all_documetns_blah} -> cursor 123
* GET /query_resutls {cursor_id: 123}

Issue #11, opened 2018-11-07
Schemas for reaction data


jayrbolton:
Add schemas that can support the example chemical reaction based gene homology query:
- genomes
- genes
- chemical_reactions
- gene_reaction_complexes

PR #12, opened 2018-11-08
(WIP) Chemical reaction gene homology schemas

jayrbolton:
I added the edges, taxa, chemical_reactions, and gene_reaction_complexes. I didn't add genes and genomes. I think @JamesJeffryes will have a better idea what should go in those.

For reference, previous schemas I wrote for genes and genomes:
https://github.com/kbase/relation_engine_spec/pull/2
https://github.com/kbase/relation_engine_spec/pull/3

Related issue https://github.com/kbase/relation_engine_spec/issues/11


schemas/vertices/chemical_reactions.json
```diff
@@ -0,0 +1,62 @@
...
+      "type": "string",
+      "example": "R03067"
+    },
+    "kegg_pathways": {
+      "type": "string",
+      "example": "Folate biosynthesis"
+    },
+    "metacyc_pathways": {
+      "type": "string",
+      "example": "AMINE-DEG|Creatinine-Degradation|Degradation"
```
JamesJeffryes: I think we probably want to update the uploader to split these | delimited fields (ids and pathways) into arrays in a future version

schemas/vertices/chemical_reactions.json
```diff
@@ -0,0 +1,62 @@
...
+    "name": {
+      "type": "string",
+      "description": "Chemical names",
+      "example": "trans-2-Methyl-5-isopropylhexa-2,5-dienal dehydrogenase_c0"
+    },
+    "gpr": {
+      "type": "string",
+      "example": "PGN_RS01070"
+    },
+    "enzyme": {
```
JamesJeffryes: I think it would be better to be more explicit and call this ec_number in future

schemas/vertices/taxa.json
```diff
@@ -0,0 +1,18 @@
...
+  "required": ["_key", "name"],
+  "properties": {
+    "_key": {
+      "type": "string",
+      "example": "haloferax_volcanii",
+      "description": "Normalized name (lower-cased and snake-cased)"
+    },
+    "name": {
+      "type": "string",
+      "example": "Haloferax Volcanii"
```
JamesJeffryes: In future I'd like to also encode the scientific lineage (class, order, family, genus, species, organism) as an array. If take from NCBI there's also a numerical id I think

schemas/edges/is_produced_by.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "required": ["_from", "_to"],
+  "description": "An entity produced from something else.",
+  "additionalProperties": true,
+  "properties": {
+    "_from": {
+      "type": "string",
+      "description": "The ID of a vertex that produces something else."
```
JamesJeffryes: I know we want to keep this generic but it would be helpful to have at least one example of the relationship (even if it's incomplete). IIRC reactions -> is_produced_by -> gene_complexes?

schemas/vertices/gene_reaction_complexes.json
```diff
@@ -0,0 +1,26 @@
...
+  "required": ["_key", "conjunctions"],
+  "properties": {
+    "_key": {
+      "type": "string",
+      "description": "Hash of the conjunctions."
+    },
+    "conjunctions": {
+      "type": "array",
+      "example": [["A", "B"], ["C", "D"]],
+      "description": "Array of conjunctions of genes. Eg: (A AND B) OR (C AND D).",
```
JamesJeffryes: I thought we were going to define these as just "AND" nodes and represent OR by having multiple complexes? I think that would make the traversal easier and it a bit closer to biological reality.

schemas/vertices/chemical_reactions.json
jayrbolton: We can do this now, might as well

schemas/vertices/chemical_reactions.json
jayrbolton: I'll change the name. What does "ec" stand for so I can put it in the description?

schemas/vertices/taxa.json
jayrbolton: We could put that in an attribute in the genome document. Or are you thinking of a separate vetex?

schemas/edges/is_produced_by.json
jayrbolton: Added example

schemas/vertices/gene_reaction_complexes.json
jayrbolton: I think I just misremembered it. Yeah the complexes are just a single flat array of AND-ed genes.

PR #13, opened 2018-11-12
Update edge schemas

JamesJeffryes:


schemas/edges/was_input_for.json
```diff
@@ -0,0 +1,23 @@
+{
```
jayrbolton: This is verbose, but can we call this `was_input_for_creation_of`?

jayrbolton: or `was_input_in_creation_of`

PR #14, opened 2018-11-13
(WIP) Add the start of the genomes and publications schemas

jayrbolton:
This is a PR into the previous PR (`reaction-schemas` branch). This is my start at adding a genome schema for feedback

I also added separte publications vertices and the `published_in` edge. 

Questions:
- do you want `ontology_events` and `ontologies_present` included? If so, separate vertices?
- I notice there are 4 separate lists for Feature, NonCodingFeature, CDS, and mRNA. Would you like to keep this distinction among vertices in the graph? Or maybe we can combine them into a single vertex that distinguishes these types?

### DIFF
--- a/schemas/edges/published_in.json
+++ b/schemas/edges/published_in.json
@@ -1,0 +1,19 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "required": ["_from", "_to"],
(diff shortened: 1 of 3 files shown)


JamesJeffryes:
So I actually pushed to the head of the branch with my changes because the genome & so I'm going to close this PR and we'll discuss in #15

PR #15, opened 2018-11-13
(WIP) Features schema

jayrbolton:
A number of attrs have a "TODO" for the description, so I'll need help on those.

This combines all DNA components (protein coding, non-coding, mrna, etc) into one vertex. I don't know if this is the best strategy. All I know is that there are a lot of overlapping attributes, so thought this may be best. I use the "type" attribute to distinguish among them.

I'm planning to use the "contains" edge to connect genomes to features. I wasn't going to put a workspace reference on these, so we have to traverse to the genome to get the workspace ID.


schemas/vertices/features.json
```diff
@@ -0,0 +1,120 @@
...
+      "type": "array",
+      "title": "Gene function descriptions",
+      "items": {"type": "string"}
+    },
+    "type": {
+      "type": "string",
+      "examples": ["Gene", "ncRNA", "repeat", "CDS", "mRNA"]
+    },
+    "location": {
+      "type": "array",
```
JamesJeffryes: Changed this up because I think we want the actual feature location rather than just an abstract start and end

schemas/vertices/publications.json
```diff
@@ -3,8 +3,20 @@
   "type": "object",
   "additionalProperties": false,
   "description": "The citation for an academic publication",
```
JamesJeffryes: Glad we are split this into its own vertex. We also want to ref this from catalog apps

schemas/vertices/features.json
```diff
@@ -0,0 +1,120 @@
...
+      "items": {"type": "string"}
+    },
+    "warnings": {
+      "type": "array",
+      "description": "Warnings generated by the uploader about this feature",
+      "items": {"type": "string"}
+    },
+    "dna_sequence": {
+      "type": "string",
+      "description": "Nucleotide sequence for this feature."
```
JamesJeffryes: This can blow up the document size. Leaving it in for now but we should consider dropping it.

schemas/vertices/features.json
```diff
@@ -0,0 +1,120 @@
...
+  "properties": {
+    "_key": {
+      "type": "string",
+      "title": "Content hash",
+      "description": "Hash of the DNA sequence for this feature."
+    },
+    "protein_translation": {
+      "type": "string",
+      "title": "Protein translation",
+      "description": "Longest coded protein (representative protein for splice variants)"
```
JamesJeffryes: Also can be big. Might want to drop this in future

schemas/vertices/features.json
```diff
@@ -0,0 +1,120 @@
...
+      "items": {"type": "string"}
+    },
+    "functional_descriptions": {
+      "type": "array",
+      "title": "Gene function descriptions",
+      "items": {"type": "string"}
+    },
+    "type": {
+      "type": "string",
+      "examples": ["Gene", "ncRNA", "repeat", "CDS", "mRNA"]
```
JamesJeffryes: I switched this to the "Type" as determined by the types in the different arrays rather than which array the feature occurs in. It's a little more helpful I thing because non_coding_features is a real grab bag of types.

schemas/vertices/features.json
jayrbolton: Maybe we can save an ID that points to a file store of a fasta?

schemas/vertices/features.json
jayrbolton: that sounds a lot better

schemas/vertices/features.json
jayrbolton: We can keep it optional and not import it for now

schemas/vertices/features.json
jayrbolton: I forgot that you mentioned a feature can be multiple contiguous segments

schemas/vertices/features.json
jayrbolton: We can leave it optional and not import any of it for now

PR #18, opened 2018-11-14
Refactor edge semantics

jayrbolton:
As discussed before, here is my proposal for generalizing edge semantics more. This way we can reuse these edges for other datasets besides the workspace. 

All queries will traverse based on "from_vertex_collection_name -> edge_name -> to_vertex_collection_name". All of these relationships can be described contextually using the set of three things.


JamesJeffryes:
I think `is_identical_to` is clearer than `is_equal_to`. Do you intend to link detail (e.g. genome) objects with this or workspace_object_versions?

`was_created_by` critically contains the object method specs for object creation so I think we need a specialized edge for that or we need to add that as an optional attribute in `was_produced_by`

I'm not sure which one of these is a replacement for `was_input_in_creation_of` I assume `was_created_with`?

To communicate usage of these edges we really need a schematic.

jayrbolton:
I was intending `was_created_with` to be used instead of `was_input_in_creation_of`.

We may be able to use arango's "graph" feature to convey how these edges get used. A graph definition is just a json array of vertex, edge, vertex triples:

```json
[ { "collection": "was_created_from", "from": [ "ws_object_versions" ], "to" : [ "ws_object_versions" ] } ]
```

Like I mentioned in slack, the downside of using generic edges is we can't have strict schemas on edge attributes, since we're reusing them in different contexts. The two options are:
* make some edges more specific (I would prefer an exact name like ws_object_was_created_by_method)
* use a generic edge but have `additionalProperties: true`

jayrbolton:
I propose we have both very generic edges and very specific edges as needed. I can create these two edges in addition to what's here:
* `ws_object_was_created_with_method`
* `ws_object_was_created_with_module`

JamesJeffryes:
I'm fine with `ws_object_was_created_with_method` is fine but shouldn't `ws_object_was_created_with_module` just use `was_created_using`

jayrbolton:
I guess that's true, I'm going in circles. I'll remove that.
A good strategy could be to start out using the highly generic edges when we can, and when we find out we need more specific stuff, make a highly specific edge.

PR #20, opened 2018-11-15
add "is_instance_of" to link objects and types (for now)

JamesJeffryes:


jayrbolton:
Looks good, I’m not crazy about the word “instance” but I’m probably being too nitpicky 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #21, opened 2018-11-15
More tweaks

JamesJeffryes:
Switched moddate over to mod_epoch. Added deleted flag. Added a few things for sdk modules

### DIFF
--- a/schemas/edges/is_consumed_by.json
+++ b/schemas/edges/is_consumed_by.json
@@ -1,0 +1,19 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "required": ["_from", "_to"],
+  "description": "An entity consumed by something else. For example a type -> is_consumed_by -> sdk_method_version",
+  "additionalProperties": true,
+  "properties": {
+    "_from": {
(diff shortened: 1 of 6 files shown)


scanon:
These look good to me so I will merge.  We can always adjust if needed.

schemas/edges/is_consumed_by.json
```diff
@@ -0,0 +1,19 @@
+{
```
jayrbolton: @JamesJeffryes My only feeback on this PR is that "consumes" is an odd word choice. To me, it means to "use up". My leaning would be something like "type_is_used_in_method" or similar.

JamesJeffryes: I chose "is_consumed_by" to pair with "is_produced_by" because I find the super long edge names clunky. I suppose I could do "is_input/output_in"

jayrbolton: Another thing I was thinking that might feel clunky/annoying now, but might be important if we later scale up to a 100 collections, is to make use of prefixes on collection names as a simple namespacing mechanism. For example, all the stuff related workspace data could have the "ws_*" prefix.

PR #22, opened 2018-11-21
(Discussion) graph definitions instead of edge schemas

jayrbolton:
This defines edges inside of whole graphs, connected to "from" and "to collections, rather than in isolation in a single schema.

The API would initialize the edges and graphs, with edge keys being taken care of automatically on creation/update.

If we decide this is a good idea, then todos are:
- [ ] update validator in this repo to validate graphs
- [ ] remove old edge schemas
- [ ] update API to initialize the database from graphs
- [ ] update the API to bulk import with validation against graph schemas

### DIFF
--- a/schemas/graphs/chemical_reactions.json
+++ b/schemas/graphs/chemical_reactions.json
@@ -1,0 +1,28 @@
+[
+  {
+    "edge": "is_similar_to",
+    "from": "reactions",
+    "to": "reactions",
+    "edgeSchema": {
+      "type": "object",
+      "required": ["jaccard_similarity"],
+      "properties": {
+        "jaccard_similarity": {
+          "type": "number",
+          "min": 0,
+          "max": 1
+        }
+      }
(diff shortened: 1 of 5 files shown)


JamesJeffryes:
I like the ability to specify graph schemas but I don't think it necessarily replaces the specification of edge schemas. If I'm not mistaken, edge collections can be used in multiple graphs and in that case I'd prefer to have the edge defined independently. We may also want the flexibility of adding vertexes and edges in ways that do not yet satisfy graph constraints (dangling edges that will eventually be satisfied in an asynchronous fashion for example).

jayrbolton:
Mainly I'm trying to scope or namespace the ege schemas better. This allows for two edges with the same name to have different schemas (ie different attributes). Say we had "similar_to" edges, one between reactions and one between genomes, and they each had required similarity score attributes with different scales and different names. How would you handle that currently?

PR #24, opened 2018-12-13
Reorganize with prefixes

jayrbolton:
This adds the prefixes "wsfull", "wsprov", "ncbi", "mash", 

I think the prefixing will work well. It assures that stuff like "ncbi_genome", "wsprov_object", and "wsfull_object" are nice and separate. It also allows us to reuse generic edge names like "contains" in different namespaces. We can also traverse across namespaces. Eg. the mash homology edges may go across ncbi_genomes and wsfull_objects

The subfolders that contains the namespace groupings can be named anything. They can be longer and more descriptive. In the future, we can put readmes in each namespace folder if we want to.


jayrbolton:
I did work on cleaning up all the old collections, but they might not be totally in sync still. We have a lot of collections already

PR #25, opened 2018-12-14
Add a couple of reaction similarity queries for KE

JamesJeffryes:


jayrbolton:
This is something I was working on avoiding but wasn't successful: `return e._to == @rid ? e._from : e._to`. Not saying we should do it now, just something to keep in mind.

I think the way to avoid it is to use a graph traversal for loop such as 

```
for rxn in rxn_reaction
  for similar_rxn edge 1..1 ANY rxn rxn_similar_to_reaction
    filter edge.df_similarity >= xyz
```
I couldn't get things working quite right though. I find queries a lot easier to write when broken up into several subqueries with `let`s. Ironically this is more of a traditional sql strategy. Seems fast enough though.

PR #30, opened 2019-01-08
Add ws provenance views

jayrbolton:
This adds views that I'm using in the workspace provenance UI.

A few things I'm noticing in this spec workflow:
- would be nice to have defaults in view params (eg. @sublink_limit defaults to 10 if not set)
- would be nice to have tests for these somehow

These queries do not yet include any access control. That can be added by adding a param for workspace ids that the user can access, and doing a filter on all object queries using this param. Should be easy. I will index the workspace ID property of the objects collection beforehand.

### DIFF
--- a/views/count_documents_in_collection.aql
+++ b/views/count_documents_in_collection.aql
@@ -1,8 +1,0 @@
-// Return count of documents in a collection
-// Args:
-//   collection - name of collection to count docs
-
-for v in @@collection
-  collect with count into length
-  return length
-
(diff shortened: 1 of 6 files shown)


JamesJeffryes:
Before I merge I want to confirm that we've not loaded any nonpublic data yet

jayrbolton:
There is no public data yet. But it may be worth adding the workspace filter to this PR before merging, since it's not hard

PR #32, opened 2019-01-09
Update rxn_gene_complex.json

JamesJeffryes:
Adds in source info (now that we'll have more than just ModelSeed).

Changed "conjuctions" to "genes" to match existing data.


jayrbolton:
Looks good to me

PR #34, opened 2019-01-14
Test access controls view and vertex

jayrbolton:
This adds a test view and vertex for the access control stuff.

### DIFF
--- a/views/list_test_vertices.aql
+++ b/views/list_test_vertices.aql
@@ -1,0 +1,8 @@
+// Test query -- lists all test vertices
+// Args:
+//   ws_ids - list of workspace ids the user has access to
+//            (this is a special param set by the relation engine API)
+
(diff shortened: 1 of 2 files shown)


views/list_test_vertices.aql
```diff
@@ -0,0 +1,8 @@
+// Test query -- lists all test vertices
+// Args:
+//   ws_ids - list of workspace ids the user has access to
+//            (this is a special param set by the relation engine API)
+
+for o in test_vertex
+  filter o.ws_id IN @ws_ids
```
JamesJeffryes: Should there be a is_public alternative?

jayrbolton: oh whoops

jayrbolton: Just added that, thanks for catching

PR #36, opened 2019-01-16
Tweak some logic in wsprov views

jayrbolton:
This checks workspace ID access on the sublinked objects and tweaks some of the query for performance using arango's explain feature

So below is the "correct" (I think) access control logic for the complicated workspace system for the gsp UI query:
- downstream references have read permissions regardless of the workspace
- upstream references have read permissions if the user has access to the workspace
- downstream provenance has read perms if the user has access to the workspace
- upstream provenance has read perms regardless of the workspace
- upstream or downstream copies have read perms if the user has access to the workspace

I found that overly complicated to implement in a query though. So instead we have kind of a subset of the above behavior in the query -- we always check for workspace access on every traversal. I think that's good enough for the short term.

### DIFF
--- a/views/wsprov_fetch_linked_objects.aql
+++ b/views/wsprov_fetch_linked_objects.aql
@@ -1,4 +1,4 @@
-// Find all linked objects to a given object
+// Find all linked objects to a given set of objects
 // Returns links of level 1, plus all child links of any nested level ("sublinks")
 // Each sublink has a "parent_id" key that points to its parent object
 // Args:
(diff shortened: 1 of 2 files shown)


views/wsprov_fetch_linked_objects.aql
```diff
@@ -8,10 +8,9 @@
 //   sublink_limit - number of sublink results (10-20 recommended)
 
 for o in wsprov_object
-  filter o._key IN @obj_keys
-  filter o.is_public || (o.workspace_id IN @ws_ids)
-  for link in 1..100 any o wsprov_links
+  filter o._key in @obj_keys
+  for link in 1..10 any o wsprov_links
```
JamesJeffryes: If I understand correctly, this effectively sets the max depth to follow links to to 10? Is this  `@sublink_limit`? Do you want to use the variable here or remove the comment on line 8?

jayrbolton: 10 is the max traversal depth. So you'd at most get links of links of links of links of links (up to 10 levels). I added this to try and reduce the query time
"sublink limit" i actually need to remove in the comments. The previous UI loaded links plus sublinks in one request. The newer UI will just load object links in one flat list. Then, if you click a link, it makes a separate request for those sublinks.

JamesJeffryes: 👍 Thx

PR #38, opened 2019-01-29
New view for collecting referencing type counts

JamesJeffryes:


jayrbolton:
looking good. what's the performance on this query?

views/list_referencing_type_counts.aql
```diff
@@ -0,0 +1,21 @@
...
+//   *** if both is_private and is_public are true this will be treated as an OR ***
+//   owners - if present, limit to objects with owner in list
+//   simplify_type - if true, strip out the module and version before collecting
+
+WITH wsprov_object
+FOR v IN 1..10
+    INBOUND CONCAT('wsprov_object/', @key) wsprov_links
+    OPTIONS {'uniqueVertices': 'global', 'bfs':true}
+    // If both is_private and is_public, return if either is true
+    FILTER (@is_private && @is_public) ? (v.is_public || v.workspace_id IN @ws_ids) :
```
jayrbolton: this could be a reason for jinja templating at some point in the future. we could remove the ternary through templating, which might speed it up

JamesJeffryes:
~800 ms thru Aardvark ~2000 ms thru RE-API

views/list_referencing_type_counts.aql
JamesJeffryes: The ternary is pretty quick, I think the main thing that would be improved is support for optional keys

jayrbolton: doing `... : (@is_private && v.workspace_id IN @ws_ids) || (@is_public && v.is_public)` might be more readable. Also, setting `is_public` to false can still show objects where `is_public` is true, I think

jayrbolton: Here's how it might work with templating:
```
{% if is_public and is_private %} FILTER v.is_public || v.workspace_id in @ws_ids {% endif %}
{% if is_public and not is_private %} FILTER v.is_public {% endif %}
{% if not is_public and is_private %} FILTER v.workspace_id in @ws_ids && !v.is_public {% endif %}
```

jayrbolton:
Seems fast.. I wonder why the API is adding 1200ms, if it's more than just network latency

views/list_referencing_type_counts.aql
JamesJeffryes: Yeah, that's how I had designed it but I could change it to exclude private or public results if the value is 0 being as it's not really optional

jayrbolton: I'm not sure I followyou here.. `is_public` can be required but still set to `false`, right?

jayrbolton:
We should consider prefixing the view name with "wsprov_"

PR #40, opened 2019-01-30
New views

JamesJeffryes:


views/wsprov_fetch_linked_objects.aql
```diff
@@ -4,12 +4,21 @@
 // Args:
 //   ws_ids - array of private workspace ids the user has access to
 //   obj_keys - wsprov_object key to find links for
+//   is_private - if present, limit to objects in workspaces that a user has access to
+//   is_public - if present, limit to objects in public workspaces
+//   *** if both is_private and is_public are true this will be treated as an OR ***
```
jayrbolton: "show_public" and "show_private" might better convey the meaning of these variables

views/wsprov_fetch_linked_objects.aql
```diff
@@ -4,12 +4,21 @@
...
-  for link in 1..10 any o wsprov_links options {uniqueVertices: 'global', bfs: true}
-    filter link.is_public || link.workspace_id IN @ws_ids
-    limit @link_limit
-    collect parent_id = o._id into links = link
-    return { links, parent_id }
+FOR o IN wsprov_object
+  FILTER o._key in @obj_keys
+  FOR v IN 1..10 ANY o wsprov_links
+    OPTIONS {uniqueVertices: 'global', bfs: true}
+    LET simple_type = SPLIT(SPLIT(v.ws_type, '-', 1)[0], '.')[1]
```
jayrbolton: Not saying we need it, but we can register a stored function for this: https://docs.arangodb.com/3.4/AQL/Extending/

views/wsprov_fetch_linked_objects.aql
```diff
@@ -4,12 +4,21 @@
...
+    FILTER (@is_private && @is_public) ? (v.is_public || v.workspace_id IN @ws_ids) :
+        (!@is_private || v.workspace_id IN @ws_ids) && (!@is_public || v.is_public)
+    FILTER (!@types || simple_type IN @types)
+    LIMIT @link_limit
+    COLLECT parent_id = o._id into objects = v
+
+    RETURN { objects, parent_id }
```
jayrbolton: Can we make a couple changes to the return value here:
- Just `RETURN objects` so that I can use pagination. Don't need to collect on the parent_id anymore
- Remove the LIMIT @link_limit -- i can limit based on the batch_size update you made

views/wsprov_fetch_linked_objects.aql
JamesJeffryes: I agree

views/wsprov_fetch_linked_objects.aql
JamesJeffryes: Interesting. seems like there are some tradeoffs with that approach WRT indexing

views/wsprov_fetch_linked_objects.aql
JamesJeffryes: I'll also convert `@object_keys` to @`key` to reflect the fact that it will work on only one ref at a time

views/wsprov_fetch_linked_objects.aql
jayrbolton: I don't think indexing would be used with an expression like that, unless we make a separate, indexed "simple_type" column. But I find that arango queries are still pretty fast without using indexes

views/wsprov_fetch_copies.aql
```diff
@@ -1,16 +1,18 @@
 // For a given object, fetch all the objects that it has been copied from or
 // to, no matter how many nested times (copies of copies of copies, forward or backward)
 // Also returns all linked objects of those copies of any nested level.
-// Each sublink result has a 'parent_id' point to the copy that it is linked from.
 // Args:
 //   ws_ids - array of private workspace ids the user has access to
-//   obj_key - key of the object (eg "1:2:3")
-//   copy_limit - max results of copies (recommended 10-20)
+//   obj_keys - wsprov_object key to find links for
```
jayrbolton: i think this should be singular

jayrbolton:
Looks good, will change the one minor typo on master. 
I wonder if that ACL filter expression can be a stored function that can be reused

PR #46, opened 2019-02-07
Update wsprov queries for UI

jayrbolton:
In my UI prototype, these are the queries I'm using (worked out using admin ad-hoc calls)

### DIFF
--- a/views/wsprov_list_referencing_type_counts.aql
+++ b/views/wsprov_list_referencing_type_counts.aql
@@ -1,18 +1,33 @@
 // Fetch the counts by type of ws_objects that refference an object
 // Args:
-//   ws_ids - array of private workspace ids the user has access to
-//   obj_key - wsprov_object key to count associations for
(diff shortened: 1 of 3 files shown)


JamesJeffryes:
The get referencing type counts query is already used by a different app. Can you register the changed query under a new name?

jayrbolton:
Yeah for sure

views/wsprov_list_referencing_type_counts.aql
```diff
@@ -30,4 +30,4 @@ LET narrs = (
     FOR v in ws_objects
     RETURN DISTINCT v.narr_name
 )
-RETURN PUSH(types, {'type': 'Narrative', 'type_count': COUNT(narrs)})
\ No newline at end of file
+RETURN PUSH(types, {'type': 'Narrative', 'type_count': COUNT(narrs)})
```
jayrbolton: My editor puts a newline char on files

PR #50, opened 2019-02-14
wsfull tweak workspaces

jayrbolton:
Adding a generic permissions edge between the user and workspace here, which mirrors the "permissions" system in the workspace itself (a/w/r/n strings). We can filter to get specific permissions.

Once this gets merged, I'll do an import on all ~40,000 prod workspaces with workspace info, usernames, and permissions


jayrbolton:
lemme know if you have any concerns. we can always tweak it.
Will go ahead and run the import in any case.

PR #55, opened 2019-02-20
Tweaks on wsfull

jayrbolton:
* I'm loosening up the additionalProperties requirement, because at this point I don't it's useful
* I want a very semantically explicit provenance edge (_from is descendant of _to, _to is ancestor of _from)

Working on a generic object importer for these four docs for prod


jayrbolton:
I agree. Mostly i want the ability to save additional columns on docs without having to do a spec PR/release while I figure out the importer

PR #56, opened 2019-02-21
Allow multiple types in wsprov_fetch_linked_objects

jayrbolton:
I lost the ability to filter by multiple types at some point. I didn't mean to lose this.


jayrbolton:
I'm okay to leave the versions as-is, i'll deal with it in the frontend

PR #57, opened 2019-03-02
View testing system

jayrbolton:
This allows spec writers to write tests for views, and to play with a live arango and RE API instance locally. Tests should run fine on travis as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,6 @@ python:
 before_script:
 - pip install jsonschema
 script:
+- docker-compose up --build -d
+- sleep 15
 - make test
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 from python:3.7-slim
 
-RUN pip install --upgrade pip requests
+RUN pip install --upgrade pip requests jsonschema
 
 COPY . /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+from python:3.7-slim
+
+RUN pip install --upgrade pip requests
+
+COPY . /app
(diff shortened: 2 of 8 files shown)


jayrbolton:
If by "toy data" you mean data that you generate in your unittests, I think that actually works very well and is the only way to do it

Otherwise you're probably talking about prototyping views and coming up with the query conceptually (ie is this query scientifically useful?). In which case I agree, toy data won't help a lot. I can't think of a non-root solution for that

JamesJeffryes:
I'm referring to the latter case but it's not just "Is this query scientifically useful?". There are also questions around "Will this query scale to our data?" 

I don't have a non-root solution for this and I see some benefit to have the specs tested here but I think it will be limited. TDD just isn't a viable solution so I think one will always be coping over a prototyped query. The tests then, are only going to catch regression errors. I'm not opposed to merging this in but I also don't think requiring unit tests of specs will get us much either.

jayrbolton:
For benchmarking, I agree we should run it on prod data, but also it would be easy to write benchmarks in python on test data that could be useful. We should also look at "explain" output more.

I gotta disagree on the other points. We don't have to require unit tests on all views. But having tests on views that get used on apps and kbase pages will add the same assurances we get writing tests for RE API endpoints, for example. It'll allow us to make non-breaking changes more easily, in order to tweak or optimize existing views. Also, currently view syntax is not validated whatsoever. We can easily write typos into the views and commit them and only find out when using through the api. I've done this a couple times now. It'd be pretty quick to write a single test that runs the view to validate syntax, params, etc. Also, some simple tests to validate auth logic in the query is pretty important I think.

PR #67, opened 2019-07-22
Reorganize queries

jayrbolton:
* Rename "views" to "stored queries"
* Move the AQL into the yaml key "query"
* Add "name" (query name) and "params" (json schema on query parameters) for each query
* Update test/validate.py to work with the above changes


test/validate.py
```diff
@@ -68,21 +68,43 @@ def validate_json_schemas():
...
+            data = yaml.safe_load(fd)
+        jsonschema.validate(data, stored_query_schema)
+        name = data['name']
         if names.get(name):
-            print('Duplicate queries named ' + name)
+            print(f'Duplicate queries named {name}')
             exit(1)
         else:
             names[name] = True
+        # Make sure `params` can be used as a JSON schema
```
MrCreosote: I'm guessing there's no easy way to check that the params match the `@` variables in the query?

test/validate.py
```diff
@@ -68,21 +68,43 @@ def validate_json_schemas():
...
+        if data.get('params'):
+            try:
+                jsonschema.validate({}, data['params'])
+            except ValidationError:
```
MrCreosote: Wait, if it fails to validate that's a pass...?

jayrbolton: All this is doing is validating that data['params'] is a valid json schema. If it fails to parse as a json schema, it will throw a SchemaError. If it's a valid json schema, then a ValidationError will get thrown on the empty object

test/validate.py
jayrbolton: We could pass the query to these endpoints:
https://www.arangodb.com/docs/stable/http/aql-query.html
We can send the query to /_api/query and it'll validate the syntax and send back the bind vars

As long as arangodb is running as part of the test runner (I believe it is), then this should work fine. I'll add this in.

test/validate.py
MrCreosote: Ok, you should add a comment for that as it's pretty confusing

jayrbolton:
@MrCreosote I added AQL syntax checking, plus bind var validation using the arangodb query parsing endpoint. Turned out to be pretty easy and uncovered a couple bugs

**Note on the @ws_ids bind var**: this is a special bind var that is set by the RE API for every query of all workspaces the requester has access to. In AQL, if you have a bind var set, but it doesn't appear in the query, then an error gets thrown. However, we don't want to use this variable in all queries (eg. taxonomy). So the RE API will prepend the clause `LET ws_ids = @ws_ids` to prevent the error. To be consistent, we can just use `ws_ids` when we need to use that var.

MrCreosote:
this is ready for re-review then?

jayrbolton:
Ready for re review

test/helpers.py
```diff
@@ -0,0 +1,32 @@
+"""
+Test helpers
+"""
+import time
+import requests
+import functools
+
+
+@functools.lru_cache(maxsize=1)
```
MrCreosote: Out of curiosity, why is this being cached?

test/helpers.py
```diff
@@ -0,0 +1,32 @@
...
+def get_config():
+    """Return configuration data for tests."""
+    return {
+        'db_url': 'http://arangodb:8529',
```
MrCreosote: Does the url need to be configurable?

test/helpers.py
```diff
@@ -0,0 +1,32 @@
...
+        'db_auth': ('root', '')
+    }
+
+
+def wait_for_arangodb():
+    """Wait for arangodb to go live."""
+    db_url = 'http://arangodb:8529'
```
MrCreosote: Same ?

test/validate.py
```diff
@@ -68,21 +77,65 @@ def validate_json_schemas():
...
+        with open(path) as fd:
+            data = yaml.safe_load(fd)
+        jsonschema.validate(data, stored_query_schema)
+        name = data['name']
         if names.get(name):
-            print('Duplicate queries named ' + name)
+            print(f'Duplicate queries named {name}')
             exit(1)
         else:
             names[name] = True
```
MrCreosote: doesn't really matter, but this could be a set

jayrbolton:
@MrCreosote I made some more updates, ready for re-review if you have time:
* Changed the `names` dicts to sets as suggested
* Moved some config into the environment / docker-compose. Makes a little more sense for that config to live there, but it doesn't need to be configurable for this repo. 
* I updated many of requirements and default values for the stored query params based on what i think makes sense with the query. Not confident this is all backwards compatible so I'll have to manually test a bunch of things

For json schema default values, I'm going to add something like this in the RE API: https://python-jsonschema.readthedocs.io/en/stable/faq/

MrCreosote:
Looks ok to me other than the question about the cached fn call. I didn't look at the stored queries though.

PR #68, opened 2019-07-26
Add taxa query with tests

jayrbolton:


test/stored_queries/test_ncbi_tax.py
```diff
@@ -0,0 +1,61 @@
...
+        'ncbi_child_of_taxon': ncbi_child_of_taxon_results
+    }
+
+
+class TestNcbiTax(unittest.TestCase):
+
+    def test_valid(self):
+        """Test a valid query."""
+        taxon_docs = [
+            {'_key': '1', 'scientific_name': 'Bacteria', 'rank': 'Domain'},
```
slebras: not sure if this is a productive comment (probably not), but using 'scientific_name' as the field here is a little confusing since it usually refers to 'Genus species'. not sure thats a big deal but thought I'd bring it up

jayrbolton: What do you mean by "genus species"?
Do you mean to say that scientific name usually refers to species name?

We're using scientific name to refer to a human-readable name assigned to any rank in the taxonomy, in contrast to the taxonomy_id

PR #69, opened 2019-07-30
Add descendants query with a couple valid test cases

jayrbolton:
Adds a descendants query for ncbi taxa

This is the first query I've tested that uses a default value in its params. I had to do a lot of tweaking of the API to get it to work, but it should work now in the API's develop branch/tag.


stored_queries/ncbi_tax/ncbi_taxon_get_descendants.yaml
```diff
@@ -0,0 +1,21 @@
...
+    levels:
+      type: integer
+      default: 1
+      title: Number of levels
+      description: How many levels of descendants to traverse and aggregate
+query: |
+  for t in ncbi_taxon
+      filter t._key == @key
+      for c in 1..@levels INBOUND t ncbi_child_of_taxon
+          return c
```
MrCreosote: Should there be a limit here? This could return a huge amount of data and maybe even oom the machine

jayrbolton:
Removed the unnecessary auth headers within these tests

jayrbolton:
Added the limit in the query (10k) with a comment about it at the top

PR #70, opened 2019-07-30
Add tax siblings query

jayrbolton:
Adds a siblings query for ncbi taxa with one minimal test

Hard limit on total results of 10k

Not much else to it

### DIFF
--- a/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
+++ b/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
@@ -10,17 +10,13 @@ params:
       title: Document key
       description: Key of the taxon vertex for which you want to find siblings
 query: |
-  // First get the id of the parent node
-  let parent_id = FIRST(
-    for t in ncbi_taxon
-      filter t._key == @key
-      for p in 1..1 OUTBOUND t ncbi_child_of_taxon
-        return p._id
-  )
-  // Get all children from the parent, filter out @key
-  for p in ncbi_taxon
-    filter p._id == parent_id
-    for c in 1..1 INBOUND p ncbi_child_of_taxon
-      filter c._key != @key
-      limit 10000
-      return c
+  // First find the taxon in the tree
+  for t in ncbi_taxon
+    filter t._key == @key
+    // Now traverse to the parent
+    for parent in 1..1 OUTBOUND t ncbi_child_of_taxon
+      // Then traverse back down to the siblings
+      for c in 1..1 INBOUND parent ncbi_child_of_taxon
+        filter c != t
+        limit 10000
+        return c
--- a/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
+++ b/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
@@ -17,7 +17,7 @@ query: |
       for p in 1..1 OUTBOUND t ncbi_child_of_taxon
         return p._id
   )
-  // Get all siblings from the parent
+  // Get all children from the parent, filter out @key
   for p in ncbi_taxon
     filter p._id == parent_id
     for c in 1..1 INBOUND p ncbi_child_of_taxon
--- a/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
+++ b/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
@@ -1,0 +1,15 @@
+# Get the array of siblings for a taxon
+name: ncbi_taxon_get_siblings
+params:
+  type: object
+  required: [key]
+  properties:
+    key:
+      type: string
+      title: Document key
+      description: Key of the taxon vertex for which you want to find descendants
+query: |
+  for t in ncbi_taxon
+      filter t._key == @key
+      for c in 1..@levels INBOUND t ncbi_child_of_taxon
+          return c
--- a/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
+++ b/stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
@@ -1,4 +1,5 @@
 # Get the array of siblings for a taxon
+# Results are limited to 10k
 name: ncbi_taxon_get_siblings
 params:
   type: object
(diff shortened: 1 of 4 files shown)


test/stored_queries/test_ncbi_tax.py
```diff
@@ -90,9 +92,31 @@ def test_descendants_2levels_valid(self):
             _CONF['re_api_url'] + '/api/v1/query_results',
             params={'stored_query': 'ncbi_taxon_get_descendants'},
             data=json.dumps({'key': '1', 'levels': 2}),
+            headers={'Authorization': 'valid_token'}  # gives access to workspaces [1,2,3]
```
MrCreosote: Is this supposed to be back?

MrCreosote: Is this supposed to be back?

test/stored_queries/test_ncbi_tax.py
```diff
@@ -90,9 +92,31 @@ def test_descendants_2levels_valid(self):
...
+            'Deltaproteobacteria',
+        })
+
+    def test_siblings_valid(self):
+        """Test a valid query for siblings."""
+        resp = requests.post(
+            _CONF['re_api_url'] + '/api/v1/query_results',
+            params={'stored_query': 'ncbi_taxon_get_siblings'},
+            data=json.dumps({'key': '5'}),  # Querying from "Alphaproteobacteria"
+            headers={'Authorization': 'valid_token'}  # gives access to workspaces [1,2,3]
```
MrCreosote: Thought this wasn't needed for tax queries

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
```diff
@@ -0,0 +1,26 @@
...
+      filter t._key == @key
+      for p in 1..1 OUTBOUND t ncbi_child_of_taxon
+        return p._id
+  )
+  // Get all children from the parent, filter out @key
+  for p in ncbi_taxon
+    filter p._id == parent_id
+    for c in 1..1 INBOUND p ncbi_child_of_taxon
+      filter c._key != @key
+      limit 10000
```
MrCreosote: I assume you're not willing to add a test for the limit

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
```diff
@@ -0,0 +1,26 @@
+# Get the array of siblings for a taxon
+# Results are limited to 10k
+name: ncbi_taxon_get_siblings
+params:
+  type: object
+  required: [key]
+  properties:
+    key:
+      type: string
```
MrCreosote: Test for bad params?

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
```diff
@@ -0,0 +1,26 @@
...
+      title: Document key
+      description: Key of the taxon vertex for which you want to find siblings
+query: |
+  // First get the id of the parent node
+  let parent_id = FIRST(
+    for t in ncbi_taxon
+      filter t._key == @key
+      for p in 1..1 OUTBOUND t ncbi_child_of_taxon
+        return p._id
```
MrCreosote: What if @key is the root

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
```diff
@@ -0,0 +1,26 @@
+# Get the array of siblings for a taxon
```
MrCreosote: Out of curiosity, how would indexes be set up for this query?

test/stored_queries/test_ncbi_tax.py
jayrbolton: Whoops. Keeps sneaking in. Removed it

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
jayrbolton: Added a test for a nonexistent key
An incorrect type on this field will be caught by the api itself

jayrbolton: Nonexistent key should probably throw a 404 or some other error. I'm not sure how to do that if it's only AQL. I think this is good enough for now

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
jayrbolton: Added a test for key being the root, in which case the results are empty

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
jayrbolton: I don't think we need to

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
jayrbolton: Here are the results of an EXPLAIN:
```
Indexes used:
 By   Type      Collection            Unique   Sparse   Selectivity   Fields        Ranges
 21   primary   ncbi_taxon            true     false       100.00 %   [ `_key` ]    (t.`_key` == "5")
  6   edge      ncbi_child_of_taxon   false    false            n/a   [ `_from` ]   base OUTBOUND
 20   primary   ncbi_taxon            true     false       100.00 %   [ `_key` ]    (p.`_id` == FIRST(#2))
 14   edge      ncbi_child_of_taxon   false    false            n/a   [ `_to` ]     base INBOUND
```

It looks like an index is used to find the doc and it parent, and then to find the children of the parent, but not to filter the key of the children.

jayrbolton: I just realized I'm splitting things out in the AQL really unnecessarily. I'm updating the query

MrCreosote: I was more asking if, say, I come up with a new query that needed a certain index, how would I make sure that index gets created?

jayrbolton: I updated the query. On one example it runs now in 180ms instead of 380ms. 
It uses the same indexes as before but avoids an extra traversal

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
MrCreosote: > An incorrect type on this field will be caught by the api itself
A test here would make sure the type is specified correctly though, so if someone accidentally changes the type def a test would fail

stored_queries/ncbi_tax/ncbi_taxon_get_siblings.yaml
jayrbolton: We don't have a setting in the config for specifying indexes, but we could add it in the future. I think we talked about this already?
For now you'd have to add the indexes via Aardvark

jayrbolton:
I haven't been able to figure out why tests are failing in travis, bleh

jayrbolton:
Okay, tests pass :man_shrugging:

PR #71, opened 2019-07-31
Add search with tests for sciname on ncbi taxa

jayrbolton:


stored_queries/ncbi_tax/ncbi_taxon_search_sci_name.yaml
```diff
@@ -0,0 +1,21 @@
+# Search for an ncbi taxon with a scientific name
+# Results are limited to 10k
```
MrCreosote: Doesn't looks like they are limited to 10k? There probably should be limits and offset and limit though, things can get real bad based on personal experience if either of those are too large

test/stored_queries/test_ncbi_tax.py
```diff
@@ -136,3 +136,52 @@ def test_siblings_nonexistent_node(self):
             data=json.dumps({'key': 'xyz'}),  # Nonexistent node
         ).json()
         self.assertEqual(resp['count'], 0)
+
+    def test_search_sciname_prefix(self):
+        """Test a query to search sciname."""
+        resp = requests.post(
+            _CONF['re_api_url'] + '/api/v1/query_results',
+            params={'stored_query': 'ncbi_taxon_search_sci_name'},
+            data=json.dumps({'search_text': 'prefix:bact'}),
```
MrCreosote: What other modifiers like this are possible? Are there any that could cause problems, given we're basically exposing this part of the arango api?

test/stored_queries/test_ncbi_tax.py
```diff
@@ -136,3 +136,52 @@ def test_siblings_nonexistent_node(self):
...
+        ).json()
+        self.assertEqual(resp['count'], 1)
```
MrCreosote: No check that the result is what you expected?

stored_queries/ncbi_tax/ncbi_taxon_search_sci_name.yaml
```diff
@@ -0,0 +1,21 @@
...
+      description: Text to search on for the scientific name
+    offset:
+      type: integer
+      default: 0
+    limit:
+      type: integer
+      default: 20
+query: |
+  FOR doc in FULLTEXT(ncbi_taxon, "scientific_name", @search_text)
+    LIMIT @offset, @limit
```
MrCreosote: Doesn't seem like these are tested

schemas/ncbi/ncbi_taxon.yaml
```diff
@@ -1,5 +1,10 @@
 name: ncbi_taxon
```
MrCreosote: So can we ditch the canonicalization of names in the nodes then?

test/stored_queries/test_ncbi_tax.py
```diff
@@ -136,3 +136,52 @@ def test_siblings_nonexistent_node(self):
...
+    def test_search_sciname_more_complicated(self):
+        """Test a query to search sciname."""
+        resp = requests.post(
+            _CONF['re_api_url'] + '/api/v1/query_results',
+            params={'stored_query': 'ncbi_taxon_search_sci_name'},
+            data=json.dumps({'search_text': "prefix:gamma,|prefix:alpha,|prefix:delta"})
+        ).json()
+        self.assertEqual(resp['count'], 3)
+        names = {r['scientific_name'] for r in resp['results']}
+        self.assertEqual(names, {'Gammaproteobacteria', 'Alphaproteobacteria', 'Deltaproteobacteria'})
```
MrCreosote: Would be good to have some tests on trickier names w/ multiple tokens, punctuation, capitalization etc

jayrbolton:
Will address your comments, thanks. I am also going to change the format of the indexes config in the schema. I realized it won't work well to have the index type as the key if we want several, separate indexes of the same type. So I will make it just an array of config objects.

jayrbolton:
Changed the format of the index config in schemas/ncbi/ncbi_taxon.yaml

schemas/ncbi/ncbi_taxon.yaml
jayrbolton: Yeah, most likely

stored_queries/ncbi_tax/ncbi_taxon_search_sci_name.yaml
jayrbolton: Changed this comment and added offset limit of 10k

test/stored_queries/test_ncbi_tax.py
jayrbolton: The prefixes are "prefix:xyz", "complete:xyz", (the default), "|xyz" (or operator), "-xyz" (negation), "+xyz" (the default). I really don't think there would be any problems exposing this

test/stored_queries/test_ncbi_tax.py
jayrbolton: Added a check on the result here

test/stored_queries/test_ncbi_tax.py
jayrbolton: I think we should hold off on this until we know that the query and data is going to be used in an actual feature. For now this is just a "demo" endpoint, so having more detailed tests won't be worth it

stored_queries/ncbi_tax/ncbi_taxon_search_sci_name.yaml
jayrbolton: Added a test for the max offset. Not going to other pagination results for now

PR #73, opened 2019-08-02
Ontology json spec

kellyhuang21:
Added version 2 of GO vertex and edge schemas


jayrbolton:
It looks like a good start, but for things to work we need to make the following changes:
* Instead of the parent directory being "schemas/edges" for both files, make the directory "schemas/GO/...". The directory underneath "schemas/" should be the namespace name (ie. "GO")
* Convert the json to YAML (you can use this https://www.json2yaml.com/)
* The schema should be nested under a top-level "schema" key. Also add top-level keys for "name" and "type". "name" should match the file name (it is the collection name), while "type" should either be "vertex" or "edge". See an example here: https://github.com/kbase/relation_engine_spec/blob/develop/schemas/gtdb/gtdb_taxon.yaml
* I would call the vertex "GO_term" instead of "GO_vertices" since "term" is more descriptive
* We will need separate edges for the terms (ie. separate schema files), such as "GO_isa", "GO_relationship", and "GO_intersection_of"

schemas/edges/GO_ontology/GO_vertices.json
```diff
@@ -0,0 +1,100 @@
...
+      "type": "string",
+      "description": "A comment for this term.",
+      "examples": ["This term was made obsolete because it refers to a class of gene products and a biological process rather than a molecular function."]
+    },
+    "subset": {
+      "type": "string",
+      "description": "This tag indicates a term subset to which this term belongs.",
+      "examples": ["goslim_pir", "goslim_plant"]
+    },
+    "synonym": {
```
MrCreosote: Does this need to be parsed further? also looks like a list

schemas/edges/GO_ontology/GO_vertices.json
```diff
@@ -0,0 +1,100 @@
...
+      "type": "string",
+      "description": "This tag describes a typed relationship between this term and another term. The value of this tag should be the relationship type id, and then the id of the target term.",
+      "examples": ["regulates GO:0006310 ! DNA recombination", "part_of GO:0007052 ! mitotic spindle organization"]
+    },
+    "is_obsolete": {
+      "type": "string",
+      "description": "Whether or not this term is obsolete.",
+      "examples": ["true"]
+    },
+    "replaced_by": {
```
MrCreosote: Edges here? At least a list

schemas/edges/GO_ontology/GO_vertices.json
```diff
@@ -0,0 +1,100 @@
...
+      "type": "string",
+      "description": "Gives a term which replaces an obsolete term.",
+      "examples": ["GO:0022609", "GO:0044848"]
+    },
+    "consider": {
```
MrCreosote: Edges here? At least a list

schemas/edges/GO_ontology/GO_vertices.json
kellyhuang21: synonyms contain typedefs after the name. I think we can leave it as is unless it would be useful to query whether a synonym is exact or related.

schemas/edges/GO_ontology/GO_vertices.json
```diff
@@ -0,0 +1,100 @@
...
+      "type": "string",
+      "description": "This tag gives a synonym for this term, some xrefs to describe the origins of the synonym, and may indicate a synonym category or scope information.",
+      "examples": ["'mitochondrial inheritance' EXACT []"]
+    },
+    "xref": {
+      "type": "string",
+      "description": "A dbxref that describes an analagous term in another vocabulary",
+      "examples": ["Wikipedia:Reproduction"]
+    },
+    "is_a": {
```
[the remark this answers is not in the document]
kellyhuang21: I think keeping this information in a node would be useful so you can can all the relationships a node has at once. I changed these to arrays.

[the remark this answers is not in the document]
MrCreosote: You can get that like this:
```
WITH ncbi_taxa_nodes FOR v, e IN 1 ANY "ncbi_taxa_nodes/1_19_06_01" ncbi_taxa_edges
    RETURN e
```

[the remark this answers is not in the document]
MrCreosote: I don't really feel strongly about this though

schemas/GO/GO_edges_consider.yaml
```diff
@@ -0,0 +1,27 @@
...
+"$schema": http://json-schema.org/draft-07/schema#
+title: GO_edges_consider
+type: edge
+description: A entry for consider edges in the Gene Ontology (GO) hierarchy
+properties:
+  id:
+    type: string
+    description: GO id
+    examples:
+    - GO:0000136__GO:0031501__consider
```
MrCreosote: If you're separating edges into collections the edge type isn't really needed as part of the id any more

PR #82, opened 2019-08-14
Add obsolete edges

kellyhuang21:
Added consider and replaced_by edges for obsolete nodes


jayrbolton:
Do you also want to throw in an "obsolete" field on the term vertices in this PR?

jayrbolton:
@kellyhuang21 you should make "is_obsolete" a boolean type rather than a string

jayrbolton:
I went ahead and did that change

PR #87, opened 2019-09-05
Update delta-related schema fields, tests, and validation

jayrbolton:
- Update ncbi taxon schemas to have delta fields (from, to, id)
- Update ncbi taxon tests to work with these fields
- Update the schema validator to more specifically validate delta vs non-delta schemas
- Update GO schemas to have the "delta: true" flag for validation

### DIFF
--- a/schemas/GO/GO_edges_consider.yaml
+++ b/schemas/GO/GO_edges_consider.yaml
@@ -1,6 +1,7 @@
----
 name: GO_edges_consider
 type: edge
+delta: true
+
 schema:
     "$schema": http://json-schema.org/draft-07/schema#
     title: GO_edges_consider
--- a/schemas/ncbi/ncbi_child_of_taxon.yaml
+++ b/schemas/ncbi/ncbi_child_of_taxon.yaml
@@ -1,15 +1,17 @@
 name: ncbi_child_of_taxon
 type: edge
+delta: true
+
 schema:
   "$schema": http://json-schema.org/draft-07/schema#
   type: object
-  required: [_from, _to, child_type]
+  required: [from, to, child_type]
   description: Edges which create the taxonomy tree for NCBI taxons.
   properties:
-    _from:
+    from:
       type: string
       description: The child. A gtdb_taxon or gtdb_organism.
-    _to:
+    to:
       type: string
       description: The parent gtdb_taxon
     child_type:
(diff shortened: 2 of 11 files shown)


schemas/GO/GO_edges_consider.yaml
```diff
@@ -1,6 +1,7 @@
----
 name: GO_edges_consider
```
MrCreosote: Note there's only 1 GO edges collection at this point.

Out of scope, but the GO and ENVO schemas are not exactly aligned to what the delta loader is producing

schemas/ncbi/ncbi_child_of_taxon.yaml
```diff
@@ -1,15 +1,17 @@
 name: ncbi_child_of_taxon
 type: edge
+delta: true
+
 schema:
   "$schema": http://json-schema.org/draft-07/schema#
   type: object
-  required: [_from, _to, child_type]
+  required: [from, to, child_type]
```
MrCreosote: add `id`

test/stored_queries/test_ncbi_tax.py
```diff
@@ -59,18 +71,18 @@ def setUpClass(cls):
             {'_from': 'ws_object_version/1:1:2', '_to': 'ncbi_taxon/1', 'assigned_by': 'assn2'},
             {'_from': 'ws_object_version/2:1:1', '_to': 'ncbi_taxon/1', 'assigned_by': 'assn2'},
         ]
-        ws_docs = [{'_key': '1', 'is_public': True}, {'_key': '2', 'is_public': False}]
+        ws_docs = [{'_key': '1', 'is_public': True}, {'_key': '1', 'is_public': False}]
```
MrCreosote: Not sure if this change was intentional? If so, what's the purpose?

schemas/ncbi/ncbi_child_of_taxon.yaml
MrCreosote: Wait, there's no `child_type` field, we took that out a while ago

test/stored_queries/test_ncbi_tax.py
jayrbolton: Good catch thanks

jayrbolton:
Re: schemas matching current data loading -- this is a job for a future PR

PR #89, opened 2019-09-11
[EXAMPLE] Update ontology query to use time-travel data

jayrbolton:


jayrbolton:
Closed in favor of #91

PR #90, opened 2019-09-13
Add preliminary schemas for genome feature nodes and edges

MrCreosote:


schemas/ws/ws_genome_features.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - feature_id
+    - workspace_id
+    - object_id
+    - version
+  properties:
+    _key:
+      type: string
+      description: The UPA and feature ID for this data
+      examples: ["35414:73:1_RSP_4039"]
+      pattern: "^\\d+:\\d+:\\d+_\\w*$"  # may need to expand the feature ID part
```
jayrbolton: To be super safe, we might do a namespace for features just in case we have another "sub-vertex" of the genome with conflicting IDs. Eg. `"\\d+:\\d+:\\d+_ft\\w*$"

schemas/ws/ws_genome_features.yaml
```diff
@@ -0,0 +1,34 @@
...
+    feature_id:
+      type: string
+      description: The unique ID of the feature within the genome
+      examples: ["RSP_4039"]
+      pattern: "^\\w*$" # may need to expand
```
jayrbolton: I'd lean towards not doing a regex here. I'm not sure how it'd benefit us and would probably slow down the schema validation.

schemas/ws/ws_genome_features.yaml
MrCreosote: Per our discussion, we'll leave as is since the collection name acts as a namespace

schemas/ws/ws_genome_features.yaml
MrCreosote: Per discussion, decided to keep this. It prevents weird characters (like space) winding up in the ID, and there's already a more complex regex on _key, so this doesn't add much more overhead. Neither of us have strong opinions though.

PR #92, opened 2019-09-13
GO schema part 1

MrCreosote:
- moved terms to new file name without any other changes. This will   make the changes easier to review in the next PR.
- deleted the old edges schemas and added the new one.

Do we need a merges schema?


jayrbolton:
Looks pretty good, no comments :+1:

PR #96, opened 2019-09-19
Add explicit field-name selection for bulk result ncbi taxa queries

jayrbolton:
Ability to select fields that get returned in the results, ala graphql

### DIFF
--- a/stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
+++ b/stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
@@ -23,6 +23,16 @@ params:
     ts:
       type: integer
       title: Versioning timestamp
+    select_obj:
+      type: array
+      items: {type: string}
+      title: WS obj fields to keep in the results
+      default: []
+    select_edge:
+      type: array
+      items: {type: string}
+      description: Taxon edge fields to keep in the results
+      default: []
 query_prefix: WITH ws_object_version
 query: |
   let count = COUNT(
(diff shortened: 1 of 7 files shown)


stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
```diff
@@ -23,6 +23,16 @@ params:
     ts:
       type: integer
       title: Versioning timestamp
+    select_obj:
+      type: array
+      items: {type: string}
+      title: WS obj fields to keep in the results
+      default: []
+    select_edge:
```
MrCreosote: I'm still pretty queasy about directly exposing database field names, but this isn't my show...

test/stored_queries/test_ncbi_tax.py
```diff
@@ -95,7 +95,7 @@ def test_get_lineage_valid(self):
         resp = requests.post(
             _CONF['re_api_url'] + '/api/v1/query_results',
             params={'stored_query': 'ncbi_taxon_get_lineage'},
-            data=json.dumps({'ts': _NOW, 'id': '7'}),
+            data=json.dumps({'ts': _NOW, 'id': '7', 'select': ['rank', 'scientific_name']}),
```
MrCreosote: So shouldn't results change here since the query is changing? Or maybe I don't understand the test?

MrCreosote:
whoops, meant to comment not approve

test/stored_queries/test_ncbi_tax.py
jayrbolton: Yeah, results will change. With no 'select', then each result will be an empty doc (these are breaking api changes). I changed the tests to select on the collection fields that i'm using the assertions below.

jayrbolton:
As we talked about in person, it doesn't seem like the collection field exposure in this pr is any greater than before, so that's a general architecture comment.

PR #98, opened 2019-09-19
Allow "select" options to be null, which selects all. Add a minimal test

jayrbolton:


MrCreosote:
Changes look fine, I personally would write a crapload more tests...

jayrbolton:
I'll expand the tests later once erik's UI is more final and he doesn't have any more requests on the queries/backend

PR #99, opened 2019-09-19
Add query to get a set of GO terms

MrCreosote:


MrCreosote:
Probably need to add an index on the `from` field for `GO_merges`

stored_queries/GO/GO_get_merges_from.yaml
```diff
@@ -0,0 +1,24 @@
...
+  required: [froms]
+  properties:
+    froms:
+      type: array
+      items:
+        type: string
+      title: GO term IDs
+      description: The list of Go term IDs from which merge edges should emanate
+      maxItems: 10000
+# It'd be nice if there was a way to get the most recent edge for each from
```
jayrbolton: Something like:
```
for o in GO_terms
  filter o.id in @froms 
  for v, e in 1..1 o GO_merges
    filter e.type == 'replaced_by'
    sort e.created desc
    limit 1
    return {from: o.id, merge: e}
```
Would probably still be pretty fast since there will be a small number of edges. Let me know if you want to do this, otherwise I'll go ahead and merge.

PR #100, opened 2019-09-19
Add GO_merges from index, _key for obj->feature edges

MrCreosote:
Will need the former for edge lookups, the latter to ensure reindexes don't
duplicate edges

Also add source field to feature -> GO edges

### DIFF
--- a/schemas/ws/ws_genome_has_feature.yaml
+++ b/schemas/ws/ws_genome_has_feature.yaml
@@ -4,8 +4,13 @@ schema:
   "$schema": http://json-schema.org/draft-07/schema#
   type: object
   description: A workspace genome has a feature.
-  required: [_from, _to]
+  required: [_from, _to, _key]
   properties:
+    _key:
+      type: string
+      examples: ['75:82:3_RSP_4039']
+      description: The unique, permanent ID of this edge. Identical to the feature _key entry.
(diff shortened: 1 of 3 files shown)


schemas/ws/ws_genome_has_feature.yaml
```diff
@@ -4,8 +4,13 @@ schema:
   "$schema": http://json-schema.org/draft-07/schema#
   type: object
   description: A workspace genome has a feature.
-  required: [_from, _to]
+  required: [_from, _to, _key]
   properties:
+    _key:
```
jayrbolton: The RE API automatically generates the _key field for edges. It creates a hash from the _from and _to keys. Is this going to be problematic for you?

MrCreosote: Actually, yeah, since I need to update the `_key` definition to include the data source. That allows other data sources to add the same edge (with potentially different fields)

MrCreosote: That's for the f->GO edges. For the edge this comment references, it probably doesn't matter.

jayrbolton: Nevermind, the RE API only auto-generates a _key for edges if the user does not supply it. So I think we're fine.

PR #104, opened 2019-09-24
Add object type information for ncbi taxon->object query

jayrbolton:
* Add a few useful fields for type vertices in the `ws` namespace
* Return the object type in the ncbi_tax_get_associated_ws_objects query

### DIFF
--- a/stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
+++ b/stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
@@ -49,9 +49,9 @@ query: |
       filter tax.created <= @ts AND tax.expired >= @ts
       limit 1
       for obj, e in 1 inbound tax ws_obj_version_has_taxon
+        filter obj.is_public or obj.workspace_id IN ws_ids
+        limit @offset, @limit
         for type in 1 outbound obj ws_obj_instance_of_type
-          filter obj.is_public or obj.workspace_id IN ws_ids
-          limit @offset, @limit
           let t = KEEP(type, ['_key', 'module_name', 'type_name', 'maj_ver', 'min_ver'])
           let o = MERGE(obj, {type: t})
           return {
--- a/stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
+++ b/stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
@@ -52,9 +52,10 @@ query: |
         for type in 1 outbound obj ws_obj_instance_of_type
           filter obj.is_public or obj.workspace_id IN ws_ids
           limit @offset, @limit
-          let o = MERGE(obj, {type: type})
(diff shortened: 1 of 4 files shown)


stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
```diff
@@ -33,7 +33,7 @@ params:
       items: {type: string}
       description: Taxon edge fields to keep in the results
       default: null
-query_prefix: WITH ws_object_version
+query_prefix: WITH ws_object_version, ws_type_version
 query: |
   let count = COUNT(
```
MrCreosote: this seems to be returning a count of all object versions attached to the taxon, including objects to which the user has no access. That seems wrong, naively.

stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
```diff
@@ -48,12 +48,15 @@ query: |
...
-      for obj, e in 1..1 inbound tax ws_obj_version_has_taxon
-        filter obj.is_public or obj.workspace_id IN ws_ids
-        limit @offset, @limit
-        return {
-          ws_obj: @select_obj ? KEEP(obj, @select_obj) : obj,
-          edge: @select_edge ? KEEP(e, @select_edge) : e
-        }
+      for obj, e in 1 inbound tax ws_obj_version_has_taxon
+        for type in 1 outbound obj ws_obj_instance_of_type
+          filter obj.is_public or obj.workspace_id IN ws_ids
```
MrCreosote: Should this filter be in the outer loop?

test/stored_queries/test_ncbi_tax.py
```diff
@@ -285,3 +296,4 @@ def test_get_associated_objs(self):
         ids = {ret['ws_obj']['_id'] for ret in results['results']}
         self.assertEqual(assignments, {'assn1', 'assn2'})
         self.assertEqual(ids, {'ws_object_version/1:1:1', 'ws_object_version/1:1:2'})
+        self.assertEqual(results['results'][0]['ws_obj']['type']['type_name'], 'Genome')
```
MrCreosote: I'd probably check the entire type structure

stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
jayrbolton: This is what the frontend/product folks requested. A little bit of a security leak, but I think negligible :man_shrugging:

stored_queries/ncbi_tax/ncbi_taxon_get_associated_ws_objects.yaml
jayrbolton: Good call, fixed

test/stored_queries/test_ncbi_tax.py
jayrbolton: Good call, fixed

PR #105, opened 2019-09-26
Fetch taxon doc from workspace object ref

jayrbolton:


jayrbolton:
More detailed tests can come later once the ui is more locked down

PR #106, opened 2019-09-27
Update feature key regex

MrCreosote:
Some features have non-word characters. Update the feature key regex to allow
any ADB allowed characters.

Also remove the regex for the feature ID so the original feature ID can
be included.


jayrbolton:
This would have been perfect for a feature branch

PR #109, opened 2019-10-03
Add new GO queries

zhlu9890:
The `is_public` filter on ws_obj makes it is difficult to find go_term_id/obj_ref for testing. So my testing is w/o the filter
I will also add `GO_get_terms_from_feature` later.


stored_queries/GO/GO_get_associated_ws_objects.yaml
```diff
@@ -0,0 +1,47 @@
...
+  FOR t in GO_terms
+    FILTER t.id == @id
+    FILTER t.created <= @ts AND t.expired >= @ts
+    limit 1
+    FOR v, e, p IN 2 INBOUND t ws_feature_has_GO_annotation, ws_genome_has_feature
+      FILTER p.edges[0].created <= @ts AND p.edges[0].expired >= @ts
+      FILTER v.is_public
+      SORT v.workspace_id ASC
+      LIMIT @offset, @limit
+      RETURN DISTINCT {
```
jayrbolton: @zhlu9890 will this return a record for every feature?

Instead can we return only a single copy of `ws_obj` and multiple copies of `feature`?

stored_queries/GO/GO_get_associated_ws_objects.yaml
```diff
@@ -0,0 +1,47 @@
...
+      maximum: 1000
+    offset:
+      type: integer
+      title: Result offset for pagination
+      default: 0
+      maximum: 100000
+    ts:
+      type: integer
+      title: Versioning timestamp
+query_prefix: ws_genome_features, ws_object_version
```
jayrbolton: Add "WITH " at the beginning of the query_prefix code

stored_queries/GO/GO_get_terms_from_ws_obj.yaml
```diff
@@ -0,0 +1,40 @@
...
+      maximum: 1000
+    offset:
+      type: integer
+      title: Result offset for pagination
+      default: 0
+      maximum: 100000
+    ts:
+      type: integer
+      title: Versioning timestamp
+query_prefix: ws_genome_features, GO_terms
```
jayrbolton: Add "WITH " at the beginning of this query_prefix code

stored_queries/GO/GO_get_associated_ws_objects.yaml
```diff
@@ -0,0 +1,47 @@
...
+query: |
+  FOR t in GO_terms
+    FILTER t.id == @id
+    FILTER t.created <= @ts AND t.expired >= @ts
+    limit 1
+    FOR v, e, p IN 2 INBOUND t ws_feature_has_GO_annotation, ws_genome_has_feature
+      FILTER p.edges[0].created <= @ts AND p.edges[0].expired >= @ts
+      FILTER v.is_public
```
jayrbolton: Change this to `v.is_public OR v.workspace_id IN ws_ids`. The variable `ws_ids` is automatically set to the api to a list of private workspaces that the user has access to

stored_queries/GO/GO_get_terms_from_ws_obj.yaml
```diff
@@ -0,0 +1,40 @@
...
+query: |
+  FOR o in ws_object_version
+    FILTER o._key == @obj_ref
+    FILTER o.is_public
```
jayrbolton: Change this to `v.is_public OR v.workspace_id IN ws_ids`. The variable `ws_ids` is automatically set to the api to a list of private workspaces that the user has access to

stored_queries/GO/GO_get_associated_ws_objects.yaml
```diff
@@ -0,0 +1,47 @@
...
+      SORT v.workspace_id ASC
+      LIMIT @offset, @limit
+      RETURN DISTINCT {
+        ws_obj: {
```
jayrbolton: Another syntax that might be easier to maintain here would be:
```
        ws_obj: KEEP(v, ['workspace_id', 'object_id', 'version', 'name']),
        feature: KEEP(p.vertices[1], ['updated_at', 'feature_id'])
```

stored_queries/GO/GO_get_terms_from_ws_feature.yaml
```diff
@@ -0,0 +1,39 @@
+# Get the terms from a feature
+
+name: GO_get_terms_from_feature
+params:
+  type: object
+  required: [feature_id, ts]
+  properties:
+    obj_ref:
+      type: string
+      title: Workspace versioned object reference
```
eapearson: this should be a feature reference now

eapearson: what am I saying? - the reference for a genome feature will need to include the object id and the feature id, right? the feature id alone is ambiguous.

stored_queries/GO/GO_get_terms_from_ws_feature.yaml
```diff
@@ -0,0 +1,39 @@
...
+    LIMIT 1
+    FOR v, e, p IN 1 OUTBOUND f ws_feature_has_GO_annotation
+      FILTER e.created <= @ts AND e.expired >= @ts
+      FILTER v.created <= @ts AND v.expired >= @ts
+      LIMIT @offset, @limit
+      RETURN DISTINCT {
+        term: v,
+        feature: {
+          feature_id: f.feature_id,
+          updated_at: f.updated_at
```
eapearson: using Jay's suggested syntax, how about
```
term: KEEP(v, 'id', 'name', 'namespace', 'alt_ids', 'def', 'comments',  'synonyms', 'xrefs', 'created', 'expired'),
feature: KEEP(p.vertices[1], 'feature_id', 'updated_at')
```
I don't think the extra term fields are necessary, but if so we can add back later. I like that the query with the return structure at least partially specified is self-documenting in terms of at least the top level of the return structure.

Also, we should just use the same field set wherever these things appear.

zhlu9890: Erik, would like me to use those fields of term for all the GO queries that return GO terms? Right now those `RETURN {term: v, edge: e}`. Also how about `edge`, do you need it raw or KEEP some fields or omit it?

stored_queries/GO/GO_get_terms_from_ws_feature.yaml
```diff
@@ -0,0 +1,44 @@
...
+      default: 0
+      maximum: 100000
+    ts:
+      type: integer
+      title: Versioning timestamp
+query_prefix: WITH GO_terms
+query: |
+  LET results=(
+    FOR f in ws_genome_features
+      FILTER f.feature_id == @feature_id
```
jayrbolton: I think we will want the input & filter to be on `f._key` rather than `f.feature_id`, which will be globally unique. I think there's enough of a risk that feature IDs will not be globally unique, so we want the genome ID prepended to make sure we don't get ambiguous results.

Also, we may need to have some auth control on this query using ws_ids, since the genomes that these features are attached to may be private. But I'm not sure about this. We can ask in standup soon

jayrbolton:
Any idea why the tests are not passing on this PR? @zhlu9890 do the tests pass locally on your machine?

jayrbolton:
I'm seeing the same failures on develop actually. I'll try to debug it there.

jayrbolton:
Fixed in develop. I think we're ready to roll now. Merging...

PR #110, opened 2019-10-03
Aliasing and subgraphs

jayrbolton:
* Add an `aliases.yaml` file that maps an alias name to collection names. This could be used by the RE API to abstract the collection names. We'll have to figure out how exactly we can use this in queries 
* Add "subgraph" configuration files. This adds another set of names that group collections together into a set of vertices and edges connecting those vertices. Examples are GO, ENVO, ncbi_taxonomy, and GTDB

Expansion (not in this PR)
* I would like to also add a concept for types on top of subgraphs, which I'd call something like "subgraph_blueprints". These would define a set of blueprint JSON schemas, collection aliases, and edge connections that a subgraph must conform to.
* For example, you might have a blueprint called "taxonomy". Both ncbi_taxonomy and gtdb would conform to this blueprint. It would define some common document properties in schemas for the "taxon" and "child_of" edges, as well as the required edge connections for the subgraph.


subgraphs/_schema.yaml
```diff
@@ -0,0 +1,27 @@
...
+    items: {type: string}
+  edges:
+    type: object
+    patternProperties:
+      # Any key
+      ".":
+        type: array
+        items:
+          type: array
+          items: [{type: string}, {type: string}]
```
MrCreosote: Descriptions of what the fields are would be helpful

subgraphs/gtdb.yaml
```diff
@@ -0,0 +1,7 @@
+id: GTDB
+name: Genome Taxonomy Database
+category: taxonomy
+url: https://gtdb.ecogenomic.org/
+vertices: [gtdb_organism, gtdb_taxon]
```
MrCreosote: Splitting this into 2 collections seems odd at first glance

test/validate.py
```diff
@@ -133,8 +134,41 @@ def validate_stored_queries():
...
+
+
+def validate_subgraphs(schemas):
+    """Validate the namespace configuration files found in /subgraphs."""
+    print('Validating subgraphs')
+    ids = set()  # type: set
+    with open('subgraphs/_schema.yaml') as fd:
+        schema = yaml.safe_load(fd)
+    for path in glob.iglob('subgraphs/**/*.yaml', recursive=True):
+        if path.endswith('_schema.yaml'):
```
MrCreosote: Probably should be == vs endswith?

aliases.yaml
```diff
@@ -0,0 +1,62 @@
+# Aliases from custom name to an array of collection names
```
MrCreosote: Why an array?

test/validate.py
```diff
@@ -133,8 +134,41 @@ def validate_stored_queries():
             _fatal((f"Bind vars are invalid.\n"
                     f"  Extra vars in query: {query_bind_vars - params}.\n"
                     f"  Extra params in schema: {params - query_bind_vars}"))
-        print(f'✓ {path} is valid.')
-    print('..all valid.')
+
+
+def validate_subgraphs(schemas):
```
MrCreosote: Document what schemas input value is

CODEOWNERS
```diff
@@ -1 +1 @@
-* @jayrbolton
+* @jayrbolton @MrCreosote
```
MrCreosote: i don't really consider myself an owner - this is your baby. I'm a contributor

aliases.yaml
jayrbolton: I imagine it could be useful an alias to expand out into multiple collections in a query

subgraphs/gtdb.yaml
jayrbolton: Removed gtdb_organism

test/validate.py
jayrbolton: Fixed

CODEOWNERS
jayrbolton: Ok, removed it.

test/validate.py
jayrbolton: Added comment about this.

PR #111, opened 2019-10-04
Update the GTDB schemas to reflect what's actually loaded.

MrCreosote:
Also update the NCBI schemas for the same reason.


jayrbolton:
As some people would say, LGTM

PR #112, opened 2019-10-10
Add query and test to fetch a taxon by scientific name

jayrbolton:
For GFU, I only want a result on exact scientific name match

### DIFF
--- a/stored_queries/ncbi_tax/ncbi_fetch_taxon_by_sciname.yaml
+++ b/stored_queries/ncbi_tax/ncbi_fetch_taxon_by_sciname.yaml
@@ -1,0 +1,18 @@
+# Fetch a taxon document by exact match on sciname
+name: ncbi_fetch_taxon_by_sciname
+params:
+  type: object
+  required: [sciname, ts]
+  properties:
+    sciname:
+      type: string
+      title: NCBI scientific name
+    ts:
+      type: integer
+      title: Versioning timestamp
+query: |
+  for t in ncbi_taxon
+      filter t.scientific_name == @sciname
+      filter t.created <= @ts AND t.expired >= @ts
+      limit 1
+      return t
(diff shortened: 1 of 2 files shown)


stored_queries/ncbi_tax/ncbi_fetch_taxon_by_sciname.yaml
```diff
@@ -0,0 +1,18 @@
...
+      type: string
+      title: NCBI scientific name
+    ts:
+      type: integer
+      title: Versioning timestamp
+query: |
+  for t in ncbi_taxon
+      filter t.scientific_name == @sciname
+      filter t.created <= @ts AND t.expired >= @ts
+      limit 1
```
MrCreosote: Hmm, are we sure sci names are unique?

jayrbolton: I'm not sure, I'm guessing it's likely they aren't unique in some cases. But it sounded to me like GFU can only get passed scientific name. It can't receive taxonomy ID (best option) or scientific name + rank. So exact match on sciname is the only option i have for moving forward with this stuff.

MrCreosote: Probably not hard to check once CI comes back up

MrCreosote: Other options are
- talk with the RAST_SDK authors to see if we can make changes so that we can pass in the taxon ID
- limit to species and strains as we talked about.

Also: https://biology.stackexchange.com/questions/73421/how-many-species-can-have-the-same-scientific-name

MrCreosote: But https://biology.stackexchange.com/questions/58075/what-instances-are-there-in-which-two-species-share-the-same-binomial-name

jayrbolton:
I'm going to merge and deploy this so it's at least available as an option for GFU for the short term.

PR #113, opened 2019-10-11
Add strain field to ncbi_taxon and rank/strain filter to sciname search query

MrCreosote:
See https://github.com/kbase/relation_engine_importers/pull/15


jayrbolton:
AFAICT LGTM
Before merging, I'm going to add a doc string on your test helper method, and move all the test helpers in this file into one place below the test class.

jayrbolton:
I also put keyword arguments on all your calls to that test helper. You might not like how it looks, but i think it'll be more maintainable. I'm releasing the specs now, but will only be able to bump the RE API to use them once CI is back up

PR #114, opened 2019-10-15
Add another test query; clean up AQL syntax; move into test subdir

jayrbolton:
I wanted an additional test query that uses a required param for use in testing bind var validation and empty results


MrCreosote:
LGTM

PR #120, opened 2019-11-01
Add a set of yaml files that describe RE import data sources for UI and documentation purposes

jayrbolton:
These files are meant to describe the data source for different subgraphs in RE (later these can correspond directly to subgraphs found in PR #110)

We can expand the contents of these files later on


data_sources/envo_ontology.yaml
```diff
@@ -0,0 +1,5 @@
+name: envo_ontology
+category: ontology
+namespace_title: Environment Ontology
+home_url: http://www.obofoundry.org/ontology/envo.html
+data_url: http://www.obofoundry.org/ontology/envo.html
```
MrCreosote: I actually got the data by trolling GitHub releases and dropping anything that wasn't valid yaml (which was most of the files).

MrCreosote: See https://github.com/kbaseIncubator/relation_engine_incubator/tree/master/relation_engine/ontologies/obograph/helper_scripts

jayrbolton: Okay I changed the data source url for Envo to be their github releases page
Their release descriptions are pretty wild

PR #125, opened 2020-04-02
Species/strain query

jayrbolton:
A simpler and hopefully faster query to be used in species/strain autocompletes in the UI

This only searches species/strain (doc.rank == "species" OR doc.strain)

### DIFF
--- a/stored_queries/taxonomy/taxonomy_search_species.yaml
+++ b/stored_queries/taxonomy/taxonomy_search_species.yaml
@@ -1,0 +1,39 @@
+# Search for a species/strain. Similar to search_sci_name, but simpler and quicker
+name: taxonomy_search_species
+params:
+  type: object
+  required: [search_text, ts, "@taxon_coll", sciname_field]
+  properties:
+    "@taxon_coll":
+      type: string
+      title: Taxon collection name
+      examples: [ncbi_taxon, gtdb_taxon]
+    search_text:
+      type: string
+      title: Search text
+      description: Text to search on for the scientific name
+    offset:
+      type: integer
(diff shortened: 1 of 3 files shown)


scanon (approved):
LGTM

stored_queries/taxonomy/taxonomy_search_species.yaml
```diff
@@ -0,0 +1,39 @@
...
+      title: Taxon collection name
+      examples: [ncbi_taxon, gtdb_taxon]
+    search_text:
+      type: string
+      title: Search text
+      description: Text to search on for the scientific name
+    offset:
+      type: integer
+      default: 0
+      maximum: 100000
```
MrCreosote, line 18: I don't think elastic will allow more than 10k will it?

MrCreosote (commented):
How fast is the search?

stored_queries/taxonomy/taxonomy_search_species.yaml
scanon: This is arango

MrCreosote: Ugh, brain fart

jayrbolton:
I have some timings I'll post in a minute

jayrbolton:
@MrCreosote Here is a sample of times for the old search_sci_name query, all in ms:
```
45.011
45.011
45.011
45.011
68.140
```
Here is a sample of times for this stripped down query (gives the same search results as above)
```
9.838
10.189
15.054
8.064
15.536
7.241
```
So it's definitely quicker

PR #129, opened 2020-04-13
adding generic ontology queries

zhlu9890:


stored_queries/generic_ontology/generic_ontology_get_ancestors.yaml
```diff
@@ -0,0 +1,43 @@
+# Get all ancestors (all parent's parents) for this term
```
jayrbolton: @zhlu9890 just to make this a little less long to read and write, can you rename the files from `generic_ontology` to just `ontology`?
For example: `generic_ontology/generic_ontology_get_ancestors.yaml` -> `ontology/ontology_get_ancestors.yaml`

jayrbolton:
I had one comment above but otherwise it looks good. Will give one more check in a few mins and then merge and release

stored_queries/generic_ontology/generic_ontology_get_ancestors.yaml
scanon: I agree with @jayrbolton.  I would rename these all and drop the generic.

scanon (changes requested):
Please rename the directory and queries to drop the "generic_" part.

zhlu9890:
updated name as requested

PR #135, opened 2020-06-17
Add Arabidopsis network queries and schemas

jayrbolton:
@ialarmedalien


schemas/djornl/djornl_domain_co_occur.yaml
```diff
@@ -0,0 +1,26 @@
...
+    A layer of protein domain co-occurrence values from the Arabidopsis
+    AraNetv2 database. The LLS scores that serve as edge values were
+    calculated from weighted mutual information scores to normalize the data
+    for comparison across studies and different types of data layers (Lee et
+    al, 2015).
+  type: object
+  required: [lls, _from, _to]
+  properties:
+    lls:
+      title: Least Likelhood Score
```
ialarmedalien: typo: Likelihood

jayrbolton: Thanks, fixed here along with several other files

schemas/djornl/djornl_pheno_assn.yaml
```diff
@@ -0,0 +1,24 @@
...
+    GWAS associations produced by analyzing a subset of phenotypes and SNPs
+    in the Arabidopsis 1001 Genomes database. Edge values are significant
+    association scores after FDR correction.
+  type: object
+  required: [assn_score, _from, _to]
+  properties:
+    assn_score:
+      title: Association Score
+      # (float)
+      type: number
```
ialarmedalien: I'm not sure how you have the db set up, but I have been dealing with this edge data as if each edge object has four properties, `_to`, `_from`, `score` (a.k.a. the edge weight), and `type` (the source). It saves having to normalise the object keys when displaying the data graphically or in tabular form. We can have a mapping of data source to what the edge weight represents. From what the Jacobson team were saying, they are interested in which nodes are connected and in the type of connection (which analysis it came from) between the nodes. `_to` and `_from` can be of any node type; again, we can have a mapping that specifies that in dataset x, `_to` is always a phenotype ID and `_from` is always a gene ID, whereas in dataset y, both are always gene IDs.

ialarmedalien: It would be useful to have a query that fetches edges filtered by `type` (e.g. `djornl_ppi_liter`, `djornl_pheno_assn`) or by `score`, plus the node data for those edges.

stored_queries/djornl/djornl_fetch_genes.yaml
```diff
@@ -0,0 +1,15 @@
...
+  properties:
+    keys:
+      type: array
+      items: {type: string}
+      title: Gene Keys
+      examples: [[AT1G01010]]
+query: |
+  FOR g IN djornl_gene
+    FILTER g._key IN @keys
+    RETURN g
```
ialarmedalien, line 15: Can you add queries to fetch node data, filtering on a search term (or terms) that appears (a) in a specified field, or (b) in any field? This can be in another PR, it doesn't have to be in this one.

schemas/djornl/djornl_pheno_assn.yaml
jayrbolton: Sure, sounds good

stored_queries/djornl/djornl_fetch_genes.yaml
jayrbolton: Arango has some good features for indexing a bunch of different fields to provide one big search endpoint ("ArangoSearch Views"). Will add a view for genes

schemas/djornl/djornl_pheno_assn.yaml
jayrbolton: Pushed an update that combines all the different edge collections into a single one called `djornl_edge` with a `score` and `score_type` field. The score_type field must be one of these strings: [domain_co_occur, gene_coexpr, pheno_assn, ppi_hithru, ppi_liter]

Also added a hash index for score_type and persistent index for score

jayrbolton:
Added 3 fields to the gene nodes for the iterative random forest clusters provided by dan:
cluster_I2
cluster_I4
cluster_I6

PR #136, opened 2020-06-18
Add djornl importer module and subdir

jayrbolton:
@ialarmedalien

### DIFF
--- a/importers/README.md
+++ b/importers/README.md
@@ -1,0 +1,20 @@
+# RE Importers
+
+This directory holds python modules that import data into ArangoDB.
+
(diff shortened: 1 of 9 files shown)


jayrbolton:
Note the base for this pr is `add-djornl-data` which is found in #135

ialarmedalien (approved):
LGTM. I would usually insist on tests, but since they may change their data format, I'll leave it.

jayrbolton:
Howabout we add a test module for the djornl importers and queries once we know this stuff is useful and used.

PR #137, opened 2020-06-23
Exascale data gene search

jayrbolton:
@ialarmedalien

Adds generic gene search with arangosearch view config

Note that this PR goes against the add-importer branch (PR #136 )

This is only tested manually (we can add a test module later). Performance is good, <40ms on my machine

### DIFF
--- a/views/djornl/djornl_gene_view.json
+++ b/views/djornl/djornl_gene_view.json
@@ -1,0 +1,82 @@
+{
+  "writebufferIdle": 64,
+  "writebufferActive": 0,
+  "type": "arangosearch",
+  "primarySort": [],
+  "writebufferSizeMax": 33554432,
+  "commitIntervalMsec": 1000,
+  "consolidationPolicy": {
+    "type": "bytes_accum",
+    "threshold": 0.1
+  },
+  "globallyUniqueId": "c98232107/",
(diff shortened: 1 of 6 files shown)


views/djornl/djornl_gene_view.json
```diff
@@ -0,0 +1,82 @@
...
+  "globallyUniqueId": "c98232107/",
+  "cleanupIntervalStep": 2,
+  "id": "98232107",
+  "links": {
+    "djornl_gene": {
+      "analyzers": [
+        "identity"
+      ],
+      "fields": {
+        "transcript": {},
```
ialarmedalien, line 21: I'd add this to the search fields, and the same with `go_terms`.

jayrbolton: By "this" do you mean the transcript field? If the value is an empty object, the default analyzer is "identity" which will still index the value but will not apply any text processing. I can add go_terms with an "identity" index, and i'll add both to the query

jayrbolton: Oh it looks like go_terms is already indexed. I just pushed a commit so that they are both in the query. Tested manually and they work.

ialarmedalien (approved):

PR #138, opened 2020-06-23
Add query to find clusters and all neighbors

jayrbolton:
@ialarmedalien

Adds query to find a gene cluster and its neighbors

Note the base for this PR is djornl-gene-search which is PR #137


ialarmedalien (approved):

jayrbolton:
FYI tweaked some stuff here to get it to work better based on manual testing

PR #140, opened 2020-06-26
Add query to fetch all verts and edges in djornl

jayrbolton:
@ialarmedalien


ialarmedalien (approved):
Amazing idea for a query! LGTM.

PR #141, opened 2020-07-08
update GO_get_associated_ws_objects.yaml to add total features count and put offset+limit on ws_obj group

zhlu9890:
to address requests from Erik on GO_get_associated_ws_objects


jayrbolton:
Looks good to me. We should add tests for these at some point. How is the performance?

PR #142, opened 2020-07-10
Exascale networking - updates

ialarmedalien:
Some updates to the stored queries, document schema, and added some tests and data to liven things up.

- The Jacobson lab have provided a new file of edge data, with all previous datasets merged into a single file. I have therefore rejigged the parser to use this file as the source.
- Also added in a bit more error checking to ensure that dodgy data doesn't make its way into Arango.
- Added parser tests and test data.

Schema changes:
- merged the phenotype and gene documents so we just refer to nodes and edges
- added docs / names / examples for all fields
- reworked the current queries to fit requirements -- all queries now return edges and nodes, with a configurable 'distance' variable


importers/djornl/parser.py
```diff
@@ -0,0 +1,213 @@
...
+                    raise RuntimeError(f"line {line_no}: expected {expected_col_count} cols, found {n_cols}")
+
+                node_ix[cols[0]] = 1
+                node_ix[cols[1]] = 1
+                edge_type = cols[4]
+                if not edge_type in edge_remap:
+                    raise RuntimeError(f"line {line_no}: invalid edge type: {edge_type}")
+
+                edges.append({
+                    '_key': f'{cols[0]}__{cols[1]}__{edge_remap[edge_type]}__{cols[2]}',
```
ialarmedalien, line 100: unique key for edges

importers/djornl/parser.py
```diff
@@ -0,0 +1,213 @@
...
+                csv_reader = csv.reader(fd, delimiter='\t')
+                for row in csv_reader:
+                    if len(row) > 1:
+                        # remove the 'Cluster' text
+                        cluster_id = row[0].replace('Cluster','')
+                        gene_keys = row[1:]
+                        nodes += [
+                            {'_key': key, cluster_label: int(cluster_id)}
+                            for key in gene_keys
+                        ]
```
ialarmedalien, line 179: Changed this slightly so that instead of having the "Cluster_n_" text repeated, we just use an integer

importers/djornl/parser.py
```diff
@@ -0,0 +1,213 @@
...
+        print(f"Saved docs to collection {coll_name}!")
+        print(resp.text)
+        print('=' * 80)
+        return resp
+
+
+    def load_data(self):
+        self.save_dataset(self.load_edges())
+        self.save_dataset(self.load_node_metadata())
+        self.save_dataset(self.load_cluster_data())
```
ialarmedalien, line 213: this is kind of ugly but it was makes testing much easier

schemas/djornl/djornl_edge.yaml
```diff
@@ -3,64 +3,45 @@ type: edge
...
+            calculated from Pearson correlation coefficients to normalize the data
+            for comparison across studies and different types of data layers (Lee et
+            al, 2015).
+        - const: pheno_assn
+          description: GWAS associations produced by analyzing a subset of phenotypes and SNPs in the Arabidopsis 1001 Genomes database. Edge values are significant association scores after FDR correction.
+        - const: ppi_hithru
+          description: Log likelihood score. A layer of protein-protein interaction values derived from four high-throughput PPI screening experiments; from the Arabidopsis AraNetv2 database. The LLS scores that serve as edge values were calculated to normalize the data for comparison across studies and different types of data layers (Lee et al, 2015).
+        - const: ppi_liter
+          description: A layer of protein-protein interaction values from literature-curated small- to medium-scale experimental data; from the Arabidopsis AraNetv2 database. The LLS scores that serve as edge values were calculated to normalize the data for comparison across studies and different types of data layers (Lee et al, 2015).
```
ialarmedalien, line 47: this is doing the same thing as the `enum`, but allows this data to be parsed

stored_queries/djornl/djornl_fetch_clusters.yaml
```diff
@@ -0,0 +1,49 @@
...
+      description: Cluster I4 IDs to locate
+      items: {type: integer}
+      examples: [[2], [4, 6]]
+      default: []
+    cluster_i6_ids:
+      title: Cluster I6 IDs
+      description: Cluster I6 IDs to locate
+      items: {type: integer}
+      examples: [[666], [999, 333]]
+      default: []
```
ialarmedalien, line 23: This is a pretty ugly way to do things -- I am still waiting to hear from the Jacobson lab about how they want to use cluster data in terms of searching, etc., but I wanted to be able to search for more than one cluster at a time. I think the best way to represent this data will probably be to have a single `clusters` field in the `node` document, and then have cluster membership listed in the form `<cluster_name>-<cluster_id>`. I will wait and see what they say first, though.

test/helpers.py
```diff
@@ -68,6 +69,37 @@ def assert_subset(testCls, subset, _dict):
...
+    # Environment variables and values to remove on exit.
+    remove_after = frozenset(k for k in update if k not in env)
+
+    try:
+        env.update(update)
+        [env.pop(k, None) for k in remove]
+        yield
+    finally:
+        env.update(update_after)
+        [env.pop(k) for k in remove_after]
```
ialarmedalien, line 100: used to modify the environment so I could test loading and parsing different file sets.

test/stored_queries/helpers.py
```diff
@@ -4,15 +4,22 @@
...
-
-def create_test_docs(coll_name, docs):
+def create_test_docs(coll_name, docs, update_on_dupe=False):
     """Create a set of documents for use in tests."""
     body = '\n'.join([json.dumps(d) for d in docs])
+    params = {'overwrite': True, 'collection': coll_name, 'display_errors': '1'}
+
+    if update_on_dupe:
+        del params['overwrite']
+        params['on_duplicate'] = 'update'
```
ialarmedalien, line 14: this duplicates the functionality of the DJORNL loader -- I did not want the test db loader to work differently to the scripts that load the Jacobson dataset.

test/stored_queries/test_djornl.py
```diff
@@ -0,0 +1,227 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+
+        # import the results file
+        results_file = os.path.join(_TEST_DIR, 'djornl', 'results.json')
+        with open(results_file) as fh:
+            lines = fh.read()
+
+        cls.json_data = json.loads(lines)
```
ialarmedalien: The results for all the stored queries were computed by my own fair hands and stored in this file.

views/djornl/djornl_node_view.json
```diff
@@ -30,11 +30,6 @@
           ]
         },
         "gene_model_type": {},
-        "mapman_desc": {
-          "analyzers": [
-            "text_en"
-          ]
-        },
```
ialarmedalien, line 37: repeated lines

views/Compounds.json
```diff
@@ -1,17 +1,16 @@
...
   "primarySort": [],
   "writebufferSizeMax": 33554432,
   "commitIntervalMsec": 1000,
   "consolidationPolicy": {
     "type": "bytes_accum",
     "threshold": 0.1
   },
-  "globallyUniqueId": "h5455DEB9D2A1/9852581",
   "cleanupIntervalStep": 10,
-  "id": "9852581",
```
ialarmedalien, line 14: unique IDs get returned by Arango when the view is created => no need to save them.

views/djornl/djornl_node_view.json
```diff
@@ -1,19 +1,19 @@
...
     "type": "bytes_accum",
     "threshold": 0.1
   },
-  "globallyUniqueId": "c98232107/",
-  "cleanupIntervalStep": 2,
-  "id": "98232107",
+  "writebufferIdle": 64,
+  "writebufferActive": 0,
+  "consolidationIntervalMsec": 60000,
+  "writebufferSizeMax": 33554432,
```
ialarmedalien, line 14: reordered lines (to match what was in the arango docs) but did not change the content of most.

importers/djornl/main.py
```diff
@@ -4,185 +4,8 @@
...
-        data='\n'.join(json.dumps(d) for d in docs)
-    )
-    if not resp.ok:
-        raise RuntimeError(resp.text)
-    else:
-        print(f"Saved docs to collection {coll_name}!")
-        print(resp.text)
-        print("=" * 80)
-
+from importers.djornl.parser import DJORNL_Parser
```
jayrbolton, line 7: Why empty out this file?

importers/djornl/parser.py
```diff
@@ -0,0 +1,213 @@
...
+"""
+import json
+import requests
+import os
+import csv
+
+import importers.utils.config as config
+
+
+class DJORNL_Parser(object):
```
jayrbolton, line 15: Why wrap everything in a class?

importers/djornl/parser.py
```diff
@@ -0,0 +1,213 @@
...
+        # AT1G01370	AT1G57820	4.40001558779779	AraNetv2_log-likelihood-score	AraNetv2-LC_lit-curated-ppi
+        edge_remap = {
+          'AraGWAS-Phenotype_Associations':         'pheno_assn',
+          'AraNetv2-CX_pairwise-gene-coexpression': 'gene_coexpr',
+          'AraNetv2-DC_domain-co-occurrence':       'domain_co_occur',
+          'AraNetv2-HT_high-throughput-ppi':        'ppi_hithru',
+          'AraNetv2-LC_lit-curated-ppi':            'ppi_liter',
+        }
+
+        node_ix = {}
```
jayrbolton, line 76: What does "ix" mean? A comment on this var would help

importers/djornl/parser.py
```diff
@@ -0,0 +1,213 @@
...
+
+                cols = [c.strip() for c in row]
+                if len(cols) != expected_col_count:
+                    n_cols = len(cols)
+                    raise RuntimeError(f"line {line_no}: expected {expected_col_count} cols, found {n_cols}")
+
+                node_ix[cols[0]] = 1
+                node_ix[cols[1]] = 1
+                edge_type = cols[4]
+                if not edge_type in edge_remap:
```
jayrbolton: You can do `edge_type not in edge_remap`. I say this mostly because flake8 complains about it for me

importers/djornl/parser.py
```diff
@@ -0,0 +1,213 @@
...
+            return self._configure()
+
+        return self._config
+
+    def _configure(self):
+
+        configuration = config.load_from_env(extra_required=['ROOT_DATA_PATH'])
+
+        # Collection name config
+        configuration['_NODE_NAME'] = 'djornl_node'
```
jayrbolton, line 28: I am a little skeptical of using one collection for both genes and phenotypes, because they are semantically totally different, and theoretically they would have totally different sets of fields, even though we use the same sets of fields for both right now. But I'm okay to leave as is, just wanted to leave a note somewhere about it.

schemas/djornl/djornl_node.yaml
```diff
@@ -1,87 +1,107 @@
-name: djornl_gene
+name: djornl_node
```
jayrbolton, line 1: I am fine to leave as-is, but want to note that I am skeptical of the idea of having one collection for both genes and phenotypes, especially if they are semantically completely different things. Also, theoretically they should have totally different sets of fields, even though we use the same set of fields for both here

stored_queries/djornl/djornl_fetch_genes.yaml
```diff
@@ -1,15 +1,37 @@
-# Fetch a gene document by list of keys
 name: djornl_fetch_genes
```
jayrbolton, line 1: Instead of repeating queries for genes and phenotypes, would you want to instead parameterize the node type?

test/stored_queries/test_djornl.py
```diff
@@ -0,0 +1,220 @@
...
+            params={'stored_query': query_name},
+            data=q_data_str
+        ).json()
+
+        return response
+
+
+    def check_expected_results(self, description, response, expected):
+
+        if _VERBOSE:
```
jayrbolton, line 80: Instead of having these conditionals everywhere we should just use the logging lib

test/stored_queries/test_djornl.py
```diff
@@ -0,0 +1,220 @@
...
+                    "fetch genes with args " + fetch_args + " and distance " + distance,
+                    resp,
+                    self.json_data['fetch_genes'][fetch_args][distance]
+                )
+
+
+    def test_fetch_clusters_no_results(self):
+
+        resp = self.submit_query('djornl_fetch_clusters', {
+            'cluster_i2_ids': [666],
```
jayrbolton, line 171: :smiling_imp:

test/validate.py
```diff
@@ -137,6 +137,46 @@ def validate_stored_queries():
     print('..all valid.')
 
 
+# JSON schema for arangosearch views found in /views
+view_schema = {
```
jayrbolton, line 141: :+1:

views/README.md
```diff
@@ -1,5 +1,3 @@
 # Views
 
-These are json files for arango views.  Currently you would need to manually create the view then cut and paste the json contents.
-
-
+These are json files for Arango views, which are required to perform searches on vertices or edges in Arango. The data in them is used by the [Relation Engine API](https://github.com/kbase/relation_engine_api) to create views via the `POST /_api/view#arangosearch` endpoint of the ArangoDB HTTP interface. Please [see the ArangoDB docs](https://www.arangodb.com/docs/3.5/http/views-arangosearch.html) for the full set of parameters available.
```
jayrbolton, line 3: You can actually still do fulltext searches without arangosearch views, but the views give you more indexing options

jayrbolton (approved):
Overall looks good and I don't have any adamant requests. I do think moving the djornl importer into a second module and wrapping it in a class are two extraneous layers. I also had some tangential thoughts about running importers, which may or may not affect your testing setup:
- importers should be able to run without the user having access to the local data and without checking in the data to the repo, meaning they would be responsible for downloading the raw data over network
- We could use either blobstore or file cache server to store the data, and when we download the data, the importer should validate a checksum
- The RE API should check whether imports have run, and if not, then kick them off. This could be an explicit poke

importers/djornl/parser.py
ialarmedalien: ix = abbreviation for index

importers/djornl/parser.py
ialarmedalien: for ease of testing, esp. with different configurations

stored_queries/djornl/djornl_fetch_genes.yaml
ialarmedalien: I was debating that; the separate queries are a legacy of having the separate collections. If there is only one type of node, it's more sensible to have a single query.

schemas/djornl/djornl_node.yaml
ialarmedalien: There are a couple of things going on here. The front end is going to display all node types in a single table, so from that perspective, it is better to have a uniform structure for nodes. As you said, we don't currently have different fields in the phenotypes and genes, and I really want to get the Jacobson group to revise their node data so that it's in a more normalised format, instead of a hodgepodge of concatenated columns and unstructured text. It was also much easier to make all the queries return results in a uniform format if we use a single table.

==> I am looking to revisit the collections after discussion with the Jacobson people.

test/stored_queries/test_djornl.py
ialarmedalien: yeah -- I should just remove them as they were for troubleshooting purposes.

ialarmedalien:
> Overall looks good and I don't have any adamant requests. I do think moving the djornl importer into a second module and wrapping it in a class are two extraneous layers. I also had some tangential thoughts about running importers, which may or may not affect your testing setup:
> 
> importers should be able to run without the user having access to the local data and without checking in the data to the repo, meaning they would be responsible for downloading the raw data over network
> We could use either blobstore or file cache server to store the data, and when we download the data, the importer should validate a checksum
> The RE API should check whether imports have run, and if not, then kick them off. This could be an explicit poke

A couple of thoughts / responses to your thoughts and responses:
- I want the Jacobson people to redo the node (gene/phenotype) metadata that they provide, and ideally give us an idea of any other data types that they're going to be using, so that we can create a more robust way of storing the data, either in a single collection or across a set of collections;
- I also want the Jacobson data to be stored somewhere that can be accessed in the same manner as the GO / ENVO / etc. ontologies, so there's no "special treatment" required for the data set. At present, it's in a GDrive folder, which is far from ideal, and I don't really want to have to write some special importer for Google Drive. Setting up something more sustainable (with version control and that is easy to access programmatically) would be much better; if their data isn't top secret, maybe a GitHub repo would work.
- I'm not attached to having the parser in a separate file and class; the reasons for moving it were (a) having an informative file name, esp. when more importers get added, and (b) making it easier to test different configurations.

importers/djornl/parser.py
jayrbolton: Okay, so it serves as a set of node keys?
My understanding is it will make sure to save a node document referenced by every edge, even if that node does not have any entry in another source file.

importers/djornl/parser.py
jayrbolton: So it's about scoping/namespacing the config, that makes sense. You can easily reinitialize some config without a class, but then you'd have to pass it around as an argument which is annoying.

PR #143, opened 2020-07-13
update GO_get_associated_ws_objects.yaml to add feature_count and add…

zhlu9890:
… GO_get_associated_ws_features.yaml

### DIFF
--- a/stored_queries/GO/GO_get_associated_ws_features.yaml
+++ b/stored_queries/GO/GO_get_associated_ws_features.yaml
@@ -1,6 +1,6 @@
-# Get the associated ws objects of this term
+# Get the associated ws features of this term
 
-name: GO_get_associated_ws_objects
+name: GO_get_associated_ws_features
 params:
   type: object
   required: [id, ts]
@@ -9,6 +9,9 @@ params:
       type: string
       title: Document ID
       description: GO id of the term you want to get the ws object of
+    obj_ref:
+      type: string
+      title: Workspace versioned object reference
     limit:
       type: integer
       title: Maximum result limit
@@ -24,6 +27,7 @@ params:
       title: Versioning timestamp
 query_prefix: WITH ws_genome_features, ws_object_version
 query: |
+  LET flag=IS_NULL(@obj_ref) OR LENGTH(@obj_ref) == 0 ? 1 : 0
   LET raw=(
     FOR t in GO_terms
       FILTER t.id == @id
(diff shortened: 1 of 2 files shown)


stored_queries/GO/GO_get_associated_ws_features.yaml
```diff
@@ -0,0 +1,49 @@
...
+      type: integer
+      title: Result offset for pagination
+      default: 0
+      maximum: 100000
+    ts:
+      type: integer
+      title: Versioning timestamp
+query_prefix: WITH ws_genome_features, ws_object_version
+query: |
+  LET flag=IS_NULL(@obj_ref) OR LENGTH(@obj_ref) == 0 ? 1 : 0
```
jayrbolton: Can you rename this to `obj_ref_null` so it's a little clearer what it represents

jayrbolton: Let this variable be a boolean rather than making it an integer in some cases. So:

```
IS_NULL(@obj_ref) OR LENGTH(@obj_ref) == 0
```

stored_queries/GO/GO_get_associated_ws_features.yaml
```diff
@@ -0,0 +1,49 @@
...
+  LET raw=(
+    FOR t in GO_terms
+      FILTER t.id == @id
+      FILTER t.created <= @ts AND t.expired >= @ts
+      limit 1
+      FOR v, e, p IN 2 INBOUND t ws_feature_has_GO_annotation, ws_genome_has_feature
+        FILTER p.edges[0].created <= @ts AND p.edges[0].expired >= @ts
+        FILTER v.is_public OR v.workspace_id IN ws_ids
+        FILTER v._key == @obj_ref OR 1 == flag
```
jayrbolton: Change to `FILTER v._key == @obj_ref OR obj_ref_null`

stored_queries/GO/GO_get_associated_ws_features.yaml
```diff
@@ -0,0 +1,49 @@
...
+        SORT v.workspace_id ASC, p.vertices[1].feature_id ASC
+        LIMIT @offset, @limit
+        RETURN DISTINCT {
```
jayrbolton, line 42: I don't think the RETURN DISTINCT will work here on nested objects. I would either try to use a COLLECT statement (this is always kind of hard to figure out), or perhaps easier would be to add the following to line 36: 

```OPTIONS {bfs: true, uniqueVertices: "global"}```

File: stored_queries/GO/GO_get_associated_ws_objects.yaml [hunk 4]
```diff
@@ -38,11 +38,7 @@ query: |
           feature: KEEP(p.vertices[1], ['feature_id', 'updated_at'])
         }
   )
-  LET total_count = COUNT(raw)
-  LET results = (
-    FOR r IN raw
-      COLLECT ws_obj=r.ws_obj INTO features=r.feature
+  FOR r IN raw
```
jayrbolton: Does it work to move this into the above query? 

```
    FOR t in GO_terms
      FILTER t.id == @id
      FILTER t.created <= @ts AND t.expired >= @ts
      LIMIT 1
      FOR v, e, p IN 2 INBOUND t ws_feature_has_GO_annotation, ws_genome_has_feature
        FILTER p.edges[0].created <= @ts AND p.edges[0].expired >= @ts
        FILTER v.is_public OR v.workspace_id IN ws_ids
        SORT v.workspace_id ASC, p.vertices[1].feature_id ASC
        COLLECT ws_obj = v WITH COUNT INTO feature_count
        RETURN DISTINCT {
          ws_obj: KEEP(ws_obj, ['workspace_id', 'object_id', 'version', 'name']),
          feature_count
        }
```

jayrbolton (changes requested):

zhlu9890:
Hi @jayrbolton, I updated PR as you suggested except the `OPTIONS {bfs: true, uniqueVertices: "global"}` which didn't give results as expected.

PR #144, opened 2020-07-16
Reorganising cluster data

ialarmedalien:
Reorganising cluster data to be a single field with an array of clusters in the form `<cluster_name>:<cluster_id>`. The current set of clusters have been renamed from 'cluster_i2' to 'markov_i2' as they were created using Markov clustering with inflation set to 2.


Makefile
```diff
@@ -1,4 +1,6 @@
 .PHONY: test
 
 test:
+	docker-compose down
 	docker-compose run spec sh /app/test/run_tests.sh
+	docker-compose down
```
ialarmedalien, line 6: make sure that any docker containers that are hanging around don't accidentally contaminate test runs.

importers/djornl/parser.py
```diff
@@ -163,24 +163,49 @@ def load_node_metadata(self):
...
+            cluster_id = row[0].replace('Cluster','')
+            node_keys = row[1:]
+
+            for key in node_keys:
+                if key not in node_ix:
+                    node_ix[key] = {}
+                if cluster_label not in node_ix[key]:
+                    node_ix[key][cluster_label] = []
+
+                node_ix[key][cluster_label].append(cluster_id)
```
ialarmedalien: makes me long for auto vivification!

test/djornl/test_data/cluster_data/out.aranetv2_subnet_AT-CX_top10percent_anno_AF_082919.abc.I2_named.tsv
```diff
@@ -1,3 +1,5 @@
+# prefix: markov_i2
+# title: Markov clustering, inflation = 2
```
ialarmedalien, line 3: I'll formalise this with the Jacobson group some time in the coming week or two. Right now, the cluster prefix is hardcoded in the parser, but ideally I'd get this information from the cluster file or from a metadata file in the clusters directory.

test/stored_queries/test_djornl.py
```diff
@@ -112,109 +113,80 @@ def test_fetch_all(self):
...
 
-    def test_fetch_phenotypes_no_results(self):
-
-        resp = self.submit_query('djornl_fetch_phenotypes', {
-            # gene node
-            "keys": ["AT1G01010"],
-        })
-        self.assertEqual(resp['results'][0], self.no_results)
-
-
```
ialarmedalien, line 129: Combined the tests for fetch calls with no results into the calls with results

test/stored_queries/test_djornl.py
```diff
@@ -112,109 +113,80 @@ def test_fetch_all(self):
...
-    @unittest.skip('This test is disabled until automated view loading is possible')
-    def test_search_nodes_no_results(self):
-
-        resp = self.submit_query('djornl_search_nodes', {
-            "search_text": "Mary Poppins",
-        })
-        self.assertEqual(resp['results'][0], self.no_results)
-
 
-    @unittest.skip('This test is disabled until automated view loading is possible')
```
ialarmedalien, line 207: enabled this test

importers/djornl/parser.py
MrCreosote: `defaultdict` might be what you're looking for: https://docs.python.org/3.8/library/collections.html

ialarmedalien: excellent, thank you @MrCreosote. That was exactly what I wanted.

schemas/djornl/djornl_node.yaml
```diff
@@ -13,21 +13,14 @@ schema:
...
-      type: integer
-      title: Cluster 4 ID
-      description: Iterative random forest cluster group ID
-      examples: [13]
-    cluster_I6:
-      type: integer
-      title: Cluster 6 ID
-      description: Iterative random forest cluster group ID
-      examples: [27]
+    clusters:
```
jayrbolton, line 20: We should add an index for this field. A hash index should be fine. The field name for the index should be "clusters[*]"

jayrbolton:
Looks good, only had one comment above

importers/djornl/manifest.schema.json
```diff
@@ -0,0 +1,52 @@
...
+          "data_type": { "enum": ["cluster"] }
+        },
+        "required": [ "prefix" ]
+      },
+      {
+        "properties": {
+          "data_type": { "enum": [ "node", "edge" ] }
+        }
+      }
+    ],
```
ialarmedalien, line 20: if the data type is "cluster", there must be a "prefix" field, which dictates the cluster label

importers/djornl/parser.py
```diff
@@ -212,3 +280,32 @@ def load_data(self):
...
+
+        # check all nodes in the edge_data have node_metadata
+        edge_no_node_md_set = edge_nodes.difference(node_metadata_nodes)
+        if edge_no_node_md_set:
+            print({'edges with no node metadata': edge_no_node_md_set})
+
+        # count all edges
+        print("Dataset contains " + str(len(edge_data['edges'])) + " edges")
+        # count all nodes
+        print("Dataset contains " + str(len(all_nodes)) + " nodes")
```
ialarmedalien, line 306: some basic sanity checks on the parsed data, plus a very high level overview of what is being added

test/stored_queries/test_djornl.py
```diff
@@ -93,128 +94,99 @@ def check_expected_results(self, description, response, expected):
...
+        self.assertEqual(
+            set(nodes_with_clusters),
+            set([json.dumps(this) for this in self.json_data['load_cluster_data']['nodes']])
+        )
 
     # indexing schema in results.json
-    # self.json_data[query][primary_param][distance_param]
-    # if primary_param is an array, join the array entities with "__"
+    # self.json_data[query_name][param_name][param_value]["distance"][distance_param]
+    # e.g. for fetch_clusters data:
```
ialarmedalien, line 115: reorganised the results data, so retrieving the appropriate results has changed further down in this file

importers/djornl/parser.py
```diff
@@ -114,73 +142,108 @@ def load_node_metadata(self):
...
-                    'mapman_desc': cols[13],
-                    'pheno_aragwas_id': cols[14],
-                    'pheno_desc1': cols[15],
-                    'pheno_desc2': cols[16],
-                    'pheno_desc3': cols[17],
-                    'pheno_ref': cols[18],
-                    'user_notes': cols[19],
-                }
-                nodes.append(doc)
+        headers = []
```
jayrbolton, line 145: Since there is a bunch of repeated parsing code between the node and edge loaders we could have a generator function for reuse:

```py

def iterate_rows(path):
    expected_col_count = 0
    with open(path) as fd:
        csv_reader = csv.reader(fd, delimiter='\t')
        line_no = 0
        for row in csv_reader:
            line_no += 1
            if len(row) <= 1 or row[0][0] == '#':
                # comment / metadata
                continue
            cols = [c.strip() for c in row]
            if line_no == 0:
                expected_col_count = len(cols)
                line_no += 1
                yield cols
            if len(cols) != expected_col_count:
                raise RuntimeError(f"{path} line {line_no}: expected {expected_col_count} cols, found {len(cols)}")
            line_no += 1
            yield cols
```
I just wrote this out without testing, but it should have the general idea.

jayrbolton (changes requested):
I had one comment, otherwise it is looking good

importers/djornl/parser.py
ialarmedalien: Yeah, I may as well make the separator also configurable/guessable from info in the manifest file.

